PR #34, opened 2026-06-12
feat: efficient agent-team model routing + bounded review workflow

### LINKED ISSUE #33: feat: efficient agent-team model routing + bounded review workflow
## Problem

Fable 5 under ultracode fans out subagents that inherit the session model. A
code review spawned 100+ Fable workers and exhausted Max-plan quota. Subagent
model resolution makes the session model the default for every worker, and the
template pinned nothing: all four role agents were `model: inherit`.

## Solution

Strong orchestrator, efficient workers, judgment roles kept strong.

- Role agents pinned in frontmatter: `developer` and `tester` run `sonnet`;
  `architect` and `reviewer` run `opus`.
- New `.claude/workflows/review-changes.js`: a bounded review. Fixed dimensions
  (bugs, security, scope, tests, style) run as Sonnet workers; one Opus critic
  consolidates. Models pinned in-script, so it never inherits an expensive
  session model and cannot fan out unboundedly (exactly N workers + 1 critic).
- `CLAUDE.md` "Model policy" section: orchestrator = Opus at max effort,
  ultracode as a per-prompt keyword, the role/model split, and the caveat that
  `CLAUDE_CODE_SUBAGENT_MODEL` must stay unset or it flattens the per-role pins.
- `sync-template` skill: add `.claude/workflows/` to the machinery copy list so
  the workflow propagates to downstream repos.

## Size

`size:S`. Config plus one workflow script plus docs.

sv-tmueller:
Closes #33

## What

- Role agents pinned in frontmatter: `developer` and `tester` = `sonnet`;
  `architect` and `reviewer` = `opus`.
- New bounded review workflow `.claude/workflows/review-changes.js`: 5 Sonnet
  dimensions (bugs, security, scope, tests, style) plus 1 Opus critic. Models
  pinned in-script, bounded by construction (exactly N workers + 1 critic).
- `CLAUDE.md`: new "Model policy" section and the `workflows/` layout line.
- `sync-template`: `.claude/workflows/` added to the machinery copy class so the
  workflow syncs to downstream repos.

## Why

Fable 5 under ultracode fans out subagents that inherit the session model (100+
Fable workers in one review). Pinning workers to Sonnet and keeping Opus for the
judgment roles gives a strong orchestrator with efficient workers.
`CLAUDE_CODE_SUBAGENT_MODEL` is deliberately left unset; it would override and
flatten the per-role pins.

## Verification

Workflow script syntax-checked as a module (parse-only, no agents spawned). The
`review-changes` workflow is being run against this diff as a dogfood; results
posted as a PR comment.

### DIFF
--- a/.claude/agents/architect.md
+++ b/.claude/agents/architect.md
@@ -2,7 +2,7 @@
 name: architect
 description: Advisory lead for approach decisions. Use when an issue needs a sub-plan before implementation, when an issue looks mis-sized or needs splitting (size:L or size:XL), or when developer and reviewer disagree. Read-only; decides approach, never writes code.
 tools: Read, Grep, Glob, Bash
-model: inherit
+model: opus
 ---
 
 You are the lead architect. You decide approach; you never write product code.
--- a/.claude/agents/developer.md
+++ b/.claude/agents/developer.md
@@ -1,7 +1,7 @@
 ---
 name: developer
 description: Implements exactly one GitHub issue end to end. Branch, TDD, conventional commits, draft PR. Use once per work package during /kickoff fan-out, for fix rounds on an existing package branch, or to implement a single refined issue on request.
-model: inherit
+model: sonnet
 isolation: worktree
 skills: superpowers:test-driven-development
 ---
--- a/.claude/agents/reviewer.md
+++ b/.claude/agents/reviewer.md
@@ -2,7 +2,7 @@
 name: reviewer
 description: Reviews a work package diff against its issue and sub-plan, in two passes, spec compliance then code quality. Read-only; outputs APPROVE or CHANGES_REQUESTED with numbered file:line findings. Never edits files.
 tools: Read, Grep, Glob, Bash
-model: inherit
+model: opus
 ---
 
 You review; you never fix. You have no Edit or Write access on purpose. Bash is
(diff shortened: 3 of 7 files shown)


sv-tmueller:
## Dogfood: ran `review-changes` on this PR

The new workflow reviewed its own diff. 6 agents (5 Sonnet reviewers + 1 Opus critic), ~126k tokens, ~2 min.

**Verdict: approve** (no must-fix). The Opus critic kept one should-fix, downgraded one finding to a nit, and dismissed two.

### should-fix (applied in a follow-up commit on this branch)

`review-changes.js`: `reviews.filter(Boolean)` silently dropped a failed reviewer's dimension while the critic was still told all 5 ran, so a dead worker (say `security`) could leave a dimension uncovered with no signal. `parallel()` null-pads on a worker error or skip per the Workflow contract, so the case is real, not speculative. Fix: track covered vs dropped dimensions by position, pass the true count to the critic, and inject a partial-coverage warning naming the uncovered dimensions when any worker fails to return. No behavior change on the happy path.

### nit (left as-is)

`REPORT_SCHEMA.required` omits `shouldFix`/`nits`, so the model may omit them. Nothing reads those keys (the report goes to the orchestrator/human), so adding them to `required` is speculative per CLAUDE.md principle 2. Revisit if a programmatic consumer appears.

### dismissed (correctly)

- Prompt injection via `args.base`: no trust boundary, since you must already be executing in the session to set args. Out of scope for a local dev tool.
- "Downstream caller breaks on missing `shouldFix`/`nits`": no such caller exists; speculative.

PR #47, opened 2026-06-13
fix: spell out three-way guard and add unknown-base fallback in sync-template

### LINKED ISSUE #39: fix: sync-template must not clobber local machinery edits
Found by the /review-codebase audit (2026-06-13). Pre-existing; not part of #37.

## What to build

The sync-template conflict guard ("update only if the local file does not differ from BOTH the old and new template versions") has no way to obtain the old template version, so it can overwrite locally modified machinery or treat every changed file as a conflict. Give it the old version so the three-way comparison works.

## Acceptance criteria

- [ ] The skill retrieves file content at the stamped template SHA (e.g. `git show <stamp>:<path>` in the template clone).
- [ ] It compares the local file against both the old and new template versions and overwrites only when the local file matches the old version.
- [ ] A locally-modified machinery file is reported as a conflict for manual merge, never silently overwritten.

sv-tmueller:
## What

Augments the Machinery bullet in `sync-template` SKILL.md Section 2 to make the three-way conflict guard executable.

## Why

The guard stated the right rule (do not overwrite if local differs from both old and new) but never explained how to obtain "the old template version". A worker reading the skill could not execute the guard without knowing the retrieval step.

## Changes

- Spells out the three versions: old = `git show <stamp>:<path>` in the template clone; new = the checked-out HEAD file; local = the file in this repo.
- Keeps the two-condition logic distinct: overwrite when local == old; conflict when local differs from both old and new; no-op when local already matches new.
- Adds the unknown-base fallback: no stamp means no old version, so the three-way test cannot run; treat machinery as a conflict, do not overwrite.
- Keeps the existing "never delete local-only agents/skills" clause verbatim.

Closes #39

### DIFF
--- a/.claude/skills/sync-template/SKILL.md
+++ b/.claude/skills/sync-template/SKILL.md
@@ -26,11 +26,16 @@ If the delta is empty, say so and stop.
 
 ## 2. File classes, in scope only
 
-- Machinery (`.claude/agents/`, `.claude/skills/`, `.claude/workflows/`): copy the template
-  version over the local file. Guard first: if the local file differs from
-  BOTH the old template version (from the stamp) and the new one, it was
-  modified locally; do not overwrite, list it in the PR as a conflict for
-  the user. Never delete local agents or skills the template does not have.
+- Machinery (`.claude/agents/`, `.claude/skills/`, `.claude/workflows/`): three-way
+  guard before copying. Get the three versions: old = `git show <stamp>:<path>` in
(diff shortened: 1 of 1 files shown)


sv-tmueller:
Tester verdict (round 1): PASS. All acceptance criteria verified against the branch; no failures. Untested / out-of-scope note: `git show <stamp>:<path>` would error for a machinery file added to the template *after* the stamp (new-file case), which is outside this issue's conflict-guard scope for modified files.

PR #49, opened 2026-06-13
fix: harden review workflow arg validation, schema, and coverage split

### LINKED ISSUE #45: chore: harden review workflows (arg validation, required schema fields, coverage message)
Found by the /review-codebase audit (2026-06-13). Pre-existing (review-changes) plus polish on review-codebase.

## What to build

Input-harden both bounded review workflows and make their report schema robust.

## Acceptance criteria

- [ ] `base` (review-changes) and `root` (review-codebase) are validated against a safe pattern (or single-quoted) before interpolation into the `git` command strings.
- [ ] `shouldFix` and `nits` are added to `REPORT_SCHEMA.required` in both workflows (or defaulted to `[]` at the consumer).
- [ ] review-codebase distinguishes scout self-drop from hard-ceiling overflow in `coverage` / `suggestedNextAction` so the caller knows which caused a shortfall.

sv-tmueller:
## Summary

Three surgical hardening fixes to `.claude/workflows/review-changes.js` and `review-codebase.js`:

- **Arg validation**: `base` (review-changes) and `root` (review-codebase) are now validated against a git-ref/path-safe char class (`[\w./\-]+`) before being interpolated into git command strings and agent prompts. Values with shell metacharacters fall back to the safe default (`origin/main` / `.`). Same `safeRef()` function shape in both files.
- **Schema required fields**: `shouldFix` and `nits` added to `REPORT_SCHEMA.required` in both files. The properties already existed; they were just optional, letting the critic omit them silently.
- **Coverage cause split** (review-codebase only): The former `scoutDropped` union is now split into `selfDrop` (scout chose not to map a path within the cap) and `overflowDrop` (script cut areas beyond `MAX_AREAS`). `ceilingReached` is derived from `overflowDrop` only. `suggestedNextAction` names the right remedy per cause (raise `args.areas` for overflow; scope `args.path` for self-drop; both if both apply). `coverageNote` in the critic's prompt is now gated on `scoutDropped.length` (the union), not only `ceilingReached`, so self-drop cases are never silently swallowed.

## Verification

- Read the git command strings: `git diff ${base}...HEAD` and `git ls-files -- ${root}` - both values are now constrained before interpolation.
- Three coverage cases traced:
  - `<= cap, empty drop`: `ceilingReached=false`, no partial note - correct
  - `> cap (overflow)`: `ceilingReached=true`, note shown, remedy = raise `args.areas` - correct
  - `<= cap, self-drop`: `ceilingReached=false` but `scoutDropped.length > 0`, note still shown, remedy = scope `args.path` - correct (this was the silent-skip bug fixed here)

Closes #45

### DIFF
--- a/.claude/workflows/review-changes.js
+++ b/.claude/workflows/review-changes.js
@@ -17,7 +17,13 @@ export const meta = {
 // Invoke with an optional base ref:
 //   Workflow({ name: 'review-changes', args: { base: 'origin/main' } })
 
-const base = (args && args.base) || 'origin/main'
+// Validate base against git-ref-safe chars; fall back to the default if it
+// contains shell metacharacters. Protects both the git command string and the
+// agent prompts that interpolate the value.
+function safeRef(value, fallback) {
+  return typeof value === 'string' && /^[\w.\/\-]+$/.test(value) ? value : fallback
+}
+const base = safeRef(args && args.base, 'origin/main')
 
 const DIMENSIONS = [
   {
@@ -87,7 +93,7 @@ const REPORT_SCHEMA = {
       description: 'findings judged false-positive or out of scope, with the reason',
     },
   },
-  required: ['verdict', 'summary', 'mustFix'],
+  required: ['verdict', 'summary', 'mustFix', 'shouldFix', 'nits'],
 }
 
 const diffHint =
--- a/.claude/workflows/review-changes.js
+++ b/.claude/workflows/review-changes.js
@@ -21,7 +21,7 @@ export const meta = {
 // contains shell metacharacters. Protects both the git command string and the
 // agent prompts that interpolate the value.
 function safeRef(value, fallback) {
-  return typeof value === 'string' && /^[\w.\/\-]+$/.test(value) ? value : fallback
+  return typeof value === 'string' && /^[\w.~^\/\-]+$/.test(value) ? value : fallback
 }
 const base = safeRef(args && args.base, 'origin/main')
 
--- a/.claude/workflows/review-codebase.js
+++ b/.claude/workflows/review-codebase.js
@@ -39,7 +39,13 @@ function parseArgs(a) {
   return {}
 }
 const opts = parseArgs(args)
-const root = opts.path || '.'
+// Validate root against path-safe chars; fall back to '.' if it contains shell
+// metacharacters. Protects both the git command string and the agent prompts
+// that interpolate the value.
+function safeRef(value, fallback) {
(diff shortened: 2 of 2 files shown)


sv-tmueller:
Tester verdict (round 1): FAIL.

Finding (must-fix): the `safeRef` regex `^[\w./\-]+$` rejects legitimate git revision syntax such as `main~3` and `main~3...HEAD` because `~` and `^` are not in the character class. The sub-plan lists `main~3...HEAD` as a value that must pass; today it silently falls back to `origin/main`. Same regex in both review-changes.js and review-codebase.js. Fix: add `~` and `^` to the allowed class, keeping shell metacharacters (`; $ ( ) backtick space newline`) rejected.

Advisory (non-blocking, out of this issue's stated shell-injection scope): `../../etc/passwd` passes the path filter for `root`; path traversal is not blocked.

sv-tmueller:
Re-test (after fix round 1): PASS. The `safeRef` regex now allows git revision operators (`~`, `^`). Verified with `node -e` against the tester's cases: all 8 legit values pass (including `main~3...HEAD` and `HEAD^`) and all 7 injection attempts (`;`, `$()`, backtick, space, newline, `|`, `&&`) are rejected. Fix pushed. Remaining advisory (out of this issue's scope, non-blocking): path traversal via `root` is still not blocked.

sv-tmueller:
Reviewer verdict (round 2): CHANGES_REQUESTED.

Must-fix: the coverage split (AC3) is inverted relative to the scout's actual behavior. The scout puts paths in `map.dropped` only when the repo EXCEEDS the ceiling (scout prompt + MAP_SCHEMA: "paths left uncovered because the repo exceeded the area ceiling"). But the code labels `map.dropped` as within-cap "self-drop" and derives `ceilingReached` solely from `allAreas.slice(MAX_AREAS)` (non-empty only if the scout DISOBEYS the cap). So in the realistic too-big-repo case the caller gets `ceilingReached: false` and a remedy that never mentions raising the cap.

Root cause: AC3 ("distinguish self-drop from ceiling overflow") assumes a within-cap self-drop the scout never performs; there is one cause (ceiling exceeded). Recommended resolution: drop the split and keep the correct single ceiling signal (combine `map.dropped` + any clamp overflow, derive `ceilingReached` from that, one remedy naming both raise-cap and scope-path) - i.e. revert AC3 to the original logic while keeping the genuinely-good hardening from this PR (arg validation, required schema fields). This narrows #45's acceptance criteria, so it needs a decision before I finish the PR.

Also: should-fix (the `suggestedNextAction` schema description still says "when ceilingReached") and a nit (`safeRef` is a misnomer for a path arg in review-codebase).

Issue #78, opened 2026-06-14
Kickoff full-pipeline runs tangle the lead's working checkout


sv-tmueller:
## Context

During `/tm-kickoff` full-pipeline runs (the `developer` agent dispatched with Agent `isolation: worktree`), the lead session's working checkout gets corrupted: after the developer returns, the main checkout is switched onto the feature branch, a detached agent worktree lingers under `.claude/worktrees/`, and a stale local feature branch is left behind (a clobber risk against the good remote branch). Seen on the #66 run and again on the parallel #60/#64 run.

#63 (PR #72) added a worktree-safety *prose* constraint to `developer.md` ("never `git switch` a shared or default checkout"), but the tangle recurred after it merged. So the cause is mechanical (the agent's `git switch -c` lands on the shared checkout despite isolation, likely a CWD / worktree-isolation interaction), not something agent wording fixes. PRs on origin are always correct; only the lead's local state is affected, and it has been cleaned up by hand after every run this session.

## What to build

A `/tm-kickoff` full-pipeline run (single package and a parallel wave) leaves the lead's checkout clean. First investigate the root cause (Agent worktree isolation vs the developer's `git switch -c` vs the dispatch CWD); then either fix the dispatch/flow so it cannot happen, or, if the cause is outside the template's control, make the kickoff skill perform a deterministic cleanup and document it.

This is an investigation: consider a short spike on the cause before committing to an approach, and re-size if it turns out deeper than `size:M`.

## Acceptance criteria

- [ ] Root cause identified and written down.
- [ ] After a single-package full-pipeline run: lead on the default branch, `git status` clean, `git worktree list` shows only the main repo, no stale local feature branch.
- [ ] The same holds after a parallel wave (2-3 packages).
- [ ] If the cause cannot be eliminated, the kickoff skill documents and performs the cleanup deterministically.
- [ ] Verified by a real `/tm-kickoff` run.

sv-tmueller:
## Sub-plan (architect, lean run)

[...]

PR #85, opened 2026-06-15
fix: stop the developer fresh path tangling the lead's checkout

sv-tmueller:
## Summary

Fixes the `/tm-kickoff` worktree tangle: after a full-pipeline run the lead's working checkout was left mutated, with a moved HEAD (switched onto the feature branch, or detached at the branch tip), sometimes a stale local feature branch, and a worktree lingering under `.claude/worktrees/`.

## Root cause

Agent `isolation: worktree` does not reliably separate the working tree for dispatched agents: their `git checkout`/`git switch` can land on the lead's shared checkout. It is harness-side, and it affects **every checkout-running agent**, not just the developer.

**Confirmed live on this run:** the `tester` dispatched for this PR ran `git checkout --detach FETCH_HEAD` and left the lead's checkout detached at the branch tip (`21a45fc`); the cleanup step below then restored it to `main`. So the original diagnosis ("only the developer's `git switch -c` tangles") was incomplete.

PR #72 added a prose warning ("never `git switch` a shared checkout"), but a worker runs the literal command it is given, so prose could not fix it.

Residue that can leak into the lead's checkout: a moved/detached HEAD (any checkout), a stale local branch (`git switch -c` specifically), and a worktree registered under `.claude/worktrees/` (harness).

## Change

Necessary-and-sufficient only in combination: eliminate the worst residue at the source, and deterministically reverse the rest (which is harness-side and unpreventable).

- **developer.md**: the fresh path is now `git checkout --detach origin/main` (or `origin/HEAD`) instead of `git switch -c`, and the existing `git push --force-with-lease origin HEAD:refs/heads/<branch>` refspec creates the branch on origin. This removes the stale **local branch** (a clobber risk against the good remote branch); it does not remove the HEAD move (a detached checkout still moves HEAD), which the cleanup handles. The fresh path also verifies the branch is absent on origin first (`git ls-remote --exit-code`) and reports `NEEDS_CONTEXT` if a prior crashed run already pushed it, so the force-push cannot silently clobber unmerged work (this replaces, and strengthens, the old local-collision clause).
- **tm-kickoff/SKILL.md**: adds a deterministic, **load-bearing** "Worktree cleanup" step. At wave end (no agents in flight) the lead re-attaches to the default branch, prunes worktrees, removes any stale `.claude/worktrees/*`, and drops a stray local package branch only when its work is safe on origin. This reverses the moved HEAD and the registered worktree the agents cannot prevent.

## Verification

Mechanical proof of the developer command change, in a throwaway sandbox:

| check | old `git switch -c` | new detached + refspec |
|---|---|---|
| local feature branch created | yes (the leak) | no |
| HEAD after the sequence | on the branch | detached |
| branch published to origin | n/a | yes |
| `main` ref | - | untouched |

The crash-work guard was also verified: `git ls-remote --exit-code` returns 0 for an existing remote branch (developer reports `NEEDS_CONTEXT`, no clobber) and non-zero when absent (fresh path proceeds).

**Live, on this run:** the tester dispatch detached the lead's checkout (the tangle reproduced itself), and the cleanup step restored it to `main` with `git worktree list` showing only the main repo. This exercises the harness-side path the sandbox cannot.

**Still deferred:** a full non-lean `/tm-kickoff` with a real worktree-isolated `developer`, single-package and a 2-3 package wave, asserting all four end-state conditions. This PR was built in lean mode (lead in a worktree), so the developer stage was not dispatched here.

Refs #78 (deliberately not `Closes`: the real-run acceptance check is deferred per "Still deferred" above, so merging this PR should not auto-close the issue).

### DIFF
--- a/.claude/agents/developer.md
+++ b/.claude/agents/developer.md
@@ -12,10 +12,10 @@ local default branch, which may be stale, so orient first:
 1. Read the issue and its comments (`gh issue view <n> --comments`). The
    sub-plan comment is your spec. If your task includes fix findings, those
    take precedence.
-2. Orient and check out, one pattern for both cases. You always work on your
-   own package branch in your dispatched worktree; never `git switch` a shared
-   or default checkout. Fetch first (`git fetch origin`), then resolve the
-   branch:
+2. Orient and check out. Both cases work on a detached HEAD in your dispatched
+   worktree and publish to your package branch on origin; you never create a
+   local branch or `git switch` a shared or default checkout. Fetch first
+   (`git fetch origin`), then resolve the base:
 
    - Fix round or resume (the branch exists on origin): verify the ref before
      detaching, so a bad branch name fails fast instead of detaching to a stale
@@ -27,15 +27,18 @@ local default branch, which may be stale, so orient first:
      ```
 
      If `ls-remote` finds no ref, stop and report `NEEDS_CONTEXT`; do not detach.
-   - Fresh package (no branch on origin): create your branch from an explicit
-     base, `git switch -c feat/<n>-<slug> origin/main` (`fix/` for bug fixes,
-     per CLAUDE.md branch naming). If the default branch is not `main`, run
+   - Fresh package (no branch on origin): check out the resolved base detached,
+     so no local branch ref is created and no shared HEAD moves. The base is
+     `origin/main`, and the eventual branch (`feat/<n>-<slug>`, or `fix/` for
+     bug fixes, per CLAUDE.md branch naming) is created on origin by the push
+     refspec below, never locally. If the default branch is not `main`, run
      `git remote set-head origin --auto` and use `origin/HEAD`; if that cannot
      resolve (for example a network error), fall back to `origin/main` and note
-     the deviation in your report. If branch creation collides with leftovers
-     from a crashed run, fetch `origin/<branch>` and inspect it: if it holds
-     work not in the upstream, stop and report `NEEDS_CONTEXT` rather than
-     discarding it; otherwise work detached from it.
+     the deviation in your report:
+
+     ```
+     git checkout --detach origin/main   # origin/HEAD when the default is not main
+     ```
 
    Publish every commit with the same refspec:
    `git push --force-with-lease origin HEAD:refs/heads/<branch>`. If no sub-plan
--- a/.claude/agents/developer.md
+++ b/.claude/agents/developer.md
@@ -27,14 +27,18 @@ local default branch, which may be stale, so orient first:
      ```
 
      If `ls-remote` finds no ref, stop and report `NEEDS_CONTEXT`; do not detach.
-   - Fresh package (no branch on origin): check out the resolved base detached,
-     so no local branch ref is created and no shared HEAD moves. The base is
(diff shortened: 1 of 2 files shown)


sv-tmueller:
## Tester verdict (round 1): FAIL

COMMIT: 21a45fc

**Finding 1 - check suite not configured.** CLAUDE.md "Useful commands" is all placeholder comments, so the tester stopped per its contract. Lead disposition: expected for this template (no suite is scaffolded by design) and the diff is documentation-only, so there is nothing to run. Treated as a SKIP, not a fix-loop trigger. This is the known "tester should emit `VERDICT: SKIP` for an unconfigured suite" nit from the 2026-06-15 review; verification here rests on the reviewer pass and the mechanical sandbox proof in the PR body.

**Finding 2 - collision guard removed; crash-work clobber (valid, fixing).** The tester is right that removing the old collision clause dropped the protection against a fresh run overwriting a prior crashed run's pushed branch. Precise framing: the primary routing guard is the lead's resume detection plus the developer's resume path (a branch on origin routes to resume, not fresh), so this is a defense-in-depth gap rather than the only guard, but it is silent data loss and worth closing. Fix: the detached fresh path now runs `git ls-remote --exit-code origin <branch>` first and reports `NEEDS_CONTEXT` if the branch already exists, which is stronger than the deleted clause (it checks the remote directly, independent of local state). Re-testing after the fix.

sv-tmueller:
## Update: root cause revised from a live observation

While running this package, the `tester` dispatch (also `isolation: worktree`) detached the lead's checkout to the branch tip `21a45fc` - the tangle reproduced itself on this very run, from the tester, not the developer. That falsifies the original "only `git switch -c` tangles" diagnosis.

Corrected root cause: the worktree isolation does not reliably separate the working tree, so **any** checkout-running agent moves the lead's HEAD. Consequences for the fix:

- The developer.md detached-checkout change still earns its place: it removes the stale **local branch** (the clobber risk), the worst residue. But it does not stop the HEAD move (a detached checkout still moves HEAD).
- The tm-kickoff cleanup is therefore **load-bearing, not a backstop**: re-attaching the lead to the default branch is what actually restores the checkout. Updated the SKILL.md wording to say so and to record the evidence.

The cleanup restored the lead to `main` cleanly when I ran it here, so the mechanism is confirmed end to end (tangle observed, then cleanup reverses it). A full non-lean run with a real worktree-isolated developer (single + wave) remains the outstanding acceptance check.

sv-tmueller:
## Reviewer verdict: APPROVE

Two passes clean. One should-fix and one nit, no must-fix.

- **should-fix (handled):** the PR linked `Closes #78`, but acceptance criterion 5 ("verified by a real `/tm-kickoff` run") and the real-run forms of criteria 2-3 are legitimately deferred (a lean run cannot dispatch the worktree-isolated developer). Changed `Closes #78` to `Refs #78` so merging does not auto-close the issue with its core acceptance check unmet; #78 stays open to track the non-lean verification.
- **nit (fixed):** the cleanup's `git switch <default>` could read as contradicting the developer's ban on `git switch`; added a half-clause noting the lead's own checkout is where that command is correct.

The reviewer confirmed the crash-work guard (tester finding 2) is adequately resolved and stronger than the deleted clause, the scope is surgical, and the writing style is clean.

PR #90, opened 2026-06-15
test: add unit tests for workflow helpers via node:vm

### LINKED ISSUE #68: Unit-test the workflow helpers
Blocked by: #59

## What to build

Automated tests for the pure helpers in the review workflows (`safeRef`, `parseArgs`, the covered/dropped partition and the `scoutDropped` union), with boundary inputs. The MAX_AREAS and `safeRef` bugs were found only by manual reasoning; a test would have caught them.

Test runner: **Node's built-in runner (`node --test` / `node:test`), zero new dependencies** (decided 2026-06-14). A `package.json` with a `test` script is the only scaffold.

Settle this design question first: can the workflow runtime `import` a shared helper module? If yes, extract the helpers and import them in both the workflows and the tests. If no, decide how to test the helpers without a tested copy drifting from the inlined real code. This is what makes the issue size:M, not a quickwin.

## Acceptance criteria

- [ ] The import question above is answered and the chosen approach noted in the PR.
- [ ] The pure helpers run under `node --test` from a documented command.
- [ ] Tests cover empty, null, "..", negative, zero, non-numeric, and valid inputs, plus the partition/union logic.
- [ ] The suite passes.

sv-tmueller:
## Summary

- Adds `package.json` (`{"private":true,"scripts":{"test":"node --test .claude/workflows/__tests__/helpers.test.mjs"}}`, zero dependencies)
- Adds `.claude/workflows/__tests__/helpers.test.mjs` with 40 tests using Node's built-in `node:test` runner
- Documents `npm test` in CLAUDE.md "Useful commands"

## What is tested

The vm approach reads each workflow `.js` as text, slices out the helper function by name, evaluates it in a `node:vm` context, and asserts against boundary inputs. Zero drift: the test runs the ACTUAL inlined code.

**1. `safeRef` (vm-loaded from tm-review-changes.js)** - 13 tests covering: valid refs pass through, empty/null/undefined/non-string fall back, `..` falls back (the bug that motivated this issue), `..` as substring falls back, shell metacharacters (`$`, `;`, backtick, `&`) fall back. Plus one assertion that the two copies in both workflow files are byte-identical.

**2. `parseArgs` (vm-loaded from tm-review-codebase.js)** - 9 tests covering: object input returns as-is, valid JSON parses, empty/whitespace string returns `{}`, malformed JSON returns `{}`, null/undefined/number return `{}`, array input returns the array (documents current behaviour).

**3. `MAX_AREAS` coercion** - 8 tests covering: valid positive integer passes through, 0/negative/non-numeric string/numeric string "5"/float/null/undefined all fall back to 24.

**4a. covered/dropped partition** - 4 tests: all present, all null, null hole at index 1, null hole at index 4.

**4b. scoutDropped union** - 6 tests: neither, self-dropped only, overflow only, both (asserts order: self-dropped first then overflow), scout failed, map.dropped is not an array.

## Honest seam note

The covered/dropped partition and scoutDropped union are inline expressions, not extractable functions. Those two groups are logic-tests: the kernel matches the source expression structure but is not evaluated from the live source. This is called out in comments at the test site.

## Test plan

- [x] `npm test` runs and all 40 tests pass (output: `pass 40, fail 0`)
- [x] No new npm dependencies introduced
- [x] Workflow `.js` files unchanged
- [x] `npm test` command documented in CLAUDE.md

Closes #68

### DIFF
--- a/.claude/workflows/__tests__/helpers.test.mjs
+++ b/.claude/workflows/__tests__/helpers.test.mjs
@@ -1,0 +1,375 @@
+/**
+ * Tests for inlined workflow helpers.
+ *
+ * Strategy: read each .js source as text, slice out the helper function by
+ * name, and evaluate it in a node:vm context. This means the test runs the
+ * ACTUAL inlined code with zero drift - no hand-copied version to maintain.
+ *
+ * Two targets are inline expressions (not extractable functions): the
+ * covered/dropped partition in tm-review-changes.js and the scoutDropped
+ * union in tm-review-codebase.js. Those are tested as logic kernels - the
+ * test captures the expression's shape, not the real source text. This is
+ * the one honest seam; it is documented below and called out in the PR.
+ */
+
+import { test, describe } from 'node:test'
+import assert from 'node:assert/strict'
+import assertLoose from 'node:assert'
+import { readFileSync } from 'node:fs'
+import { createContext, runInContext } from 'node:vm'
+import { fileURLToPath } from 'node:url'
+import { join, dirname } from 'node:path'
+
+const __dir = dirname(fileURLToPath(import.meta.url))
+const workflowsDir = join(__dir, '..')
+
+// ---------------------------------------------------------------------------
+// Helper: slice a named function out of a JS source file and evaluate it in a
+// vm context. Returns the function value.
+// ---------------------------------------------------------------------------
+function loadFn(filename, fnName) {
+  const src = readFileSync(join(workflowsDir, filename), 'utf8')
+  // Match "function <fnName>(" - handles single-line bodies too.
+  const startRe = new RegExp(`function ${fnName}\\s*\\(`)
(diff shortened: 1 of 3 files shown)


sv-tmueller:
## Tester verdict: PASS

First real suite run in this repo: `npm test` -> `node --test` = **40/40 pass, exit 0**. The tester confirmed the tests vm-load the REAL inlined `safeRef`/`parseArgs` source (byte-identical, no hand-copy); the `..` path-traversal case is load-bearing (it fails if the guard is removed); the cross-file identity assertion guards the known duplicate; the `MAX_AREAS`/partition/union logic-kernel tests are honestly flagged as reproduced kernels; all sub-plan boundary inputs are covered; no new dependencies; and the diff is confined to `package.json`, the test file, and CLAUDE.md. No findings.

Issue #130, opened 2026-06-30
Batch: package-team-as-plugin


sv-tmueller:
Tracking issue for the advisor batch that packages the orchestrator team as an
installable Claude Code plugin, distributed via a git marketplace and installed
on another Claude Code machine.

This is the approved contract. Decisions made during the run, parked questions,
and the final report are appended as comments.

## Approved proposal (contract)

One package, `size:M` (with L-risk, see sizing note).

### Package 1 - Package the orchestrator team as an installable Claude Code plugin

**Scope.** Add a Claude Code plugin manifest (`.claude-plugin/plugin.json`) and a
single-plugin marketplace catalog (`.claude-plugin/marketplace.json`) to this
template repo, so the whole team (4 agents, the tm- skills, both review
workflows, `team-guide.md`, default settings) installs on another Claude Code
machine with `/plugin marketplace add` then `/plugin install`, no hand-copying.
The manifest points its component paths at the existing `.claude/` directories
instead of moving files, so the repo keeps working as a template and as a
committed-team repo. The package includes a real install test through the
git-marketplace path and install docs. The plugin is a third deployment model
alongside the existing template-repo and `/tm-install-team` paths, not a
replacement.

**Acceptance criteria** (written to be deliverable whichever way the two
unknowns resolve):

1. `plugin.json` and `marketplace.json` exist; the repo installs as a plugin via
   the marketplace path, verified in a throwaway `CLAUDE_CONFIG_DIR` and cleaned
   up after (because `/plugin install` writes to `~/.claude/plugins/cache`, which
   a worktree does not isolate). `--plugin-dir` is the fast iteration loop.
2. After install, the 4 agents and the tm- skills load and invoke under the
   plugin namespace.
3. The two review workflows (tm-review-changes, tm-review-codebase) either
   register from the plugin or ship with a documented, working invocation (a thin
   skill that calls the script at `${CLAUDE_PLUGIN_ROOT}/workflows/...`).
4. Skill/agent references that would break invocation when installed outside this
   repo are fixed; purely informational design-doc pointers may remain and are
   listed as known limitations.
5. The superpowers prerequisite is documented (a plugin cannot auto-install
   another plugin), mirroring the instruct step `/tm-install-team` already does.
6. README gains the plugin install path as a third option.

**Size:** `M`, with explicit L-risk. The same skill files now serve three
deployment models with different roots (committed `.claude/...`, config-dir
[...]

PR #132, opened 2026-06-30
feat: package the orchestrator team as an installable Claude Code plugin

### LINKED ISSUE #131: feat: package the orchestrator team as an installable Claude Code plugin
Part of batch #130

Package the orchestrator team in this template repo as an installable Claude
Code plugin, distributed via a git marketplace, so it can be installed on
another Claude Code machine with `/plugin marketplace add` then `/plugin
install`, no hand-copying.

## Scope

Add a Claude Code plugin manifest (`.claude-plugin/plugin.json`) and a
single-plugin marketplace catalog (`.claude-plugin/marketplace.json`) to this
repo. The manifest points its component paths at the existing `.claude/`
directories instead of moving files, so the repo keeps working as a template and
as a committed-team repo. Include a real install test through the git-marketplace
path and install docs. The plugin is a third deployment model alongside the
existing template-repo and `/tm-install-team` paths, not a replacement.

The whole team goes in: the 4 agents, the tm- skills, both review workflows,
`team-guide.md`, and default settings.

## Acceptance criteria

1. `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` exist; the
   repo installs as a plugin via the marketplace path, verified in a throwaway
   `CLAUDE_CONFIG_DIR` and cleaned up after (because `/plugin install` writes to
   `~/.claude/plugins/cache`, which a worktree does not isolate). Use
   `--plugin-dir` as the fast iteration loop.
2. After install, the 4 agents and the tm- skills load and invoke under the
   plugin namespace `quinnslab` (for example `/quinnslab:tm-advisor`).
3. The two review workflows (tm-review-changes, tm-review-codebase) either
   register from the plugin or ship with a documented, working invocation (a thin
   skill that calls the script at `${CLAUDE_PLUGIN_ROOT}/workflows/...`). Verify
   workflow registration first; fall back to the wrapper skill.
4. Skill/agent references that would break invocation when installed outside this
   repo are fixed (hard-coded `/tm-...` invocations that need the namespace; any
   file a skill actually depends on reading). Purely informational design-doc
   pointers may remain and must be listed as known limitations in the PR.
5. The superpowers prerequisite is documented (a plugin cannot auto-install
   another plugin), mirroring the instruct step `/tm-install-team` already does.
6. README gains the plugin install path as a third option, alongside the existing
   template-repo and `/tm-install-team` options.

## Names (decided)

- Plugin name: `quinnslab`
- Marketplace name: `claude-template`
- Install: `/plugin install quinnslab@claude-template`
- Commands: `/quinnslab:tm-advisor`, `/quinnslab:tm-kickoff`, etc.

## Implementation notes

- Keep the files shared via manifest path-override; do not move or duplicate the
  `.claude/` tree. The same skill files serve three deployment roots (committed
  `.claude/...`, config-dir `<config>/...`, plugin `${CLAUDE_PLUGIN_ROOT}/...`);
  no single reference form is correct in all three, so fix only invocation-
  breaking references in this package.
- `--plugin-dir ./` (or the repo root) is the iteration loop; it should namespace
  the skills without polluting the global config.
- Validate `marketplace.json` resolution via a throwaway `CLAUDE_CONFIG_DIR`,
  since `/plugin install` writes to the global cache. Clean up after.

## Non-goals

- Publishing to a public marketplace, or npm distribution.
- The Claude.ai desktop chat app (cannot run this tooling).
- Moving or restructuring the existing `.claude/` layout.
- Rewriting every informational doc pointer, or forking a separately maintained
  plugin copy of the skills (a batch-2 decision if needed).
- Touching or replacing `/tm-install-team` or the template-repo flow.

## Size

`size:M`, with L-risk. If making the shared skill files behave correctly across
all three deployment models proves large, the path-reference hardening splits
into a batch-2 package rather than bloating this one.

sv-tmueller:
Closes #131

Packages the orchestrator team as an installable Claude Code plugin
(`sv-tmueller@claude-template`). Round 2 (this state): the plugin root moved
to `.claude/` so default-directory discovery registers the agents; round 1's
`source: "./"` plus explicit `agents`/`skills` override arrays left
`claude plugin details` reporting `Agents (0)`.

## What this adds

- `.claude-plugin/marketplace.json` (repo root) - single-plugin catalog,
  `source: "./.claude"`. The plugin root is `.claude/`, not the repo root.
- `.claude/.claude-plugin/plugin.json` - minimal manifest: `name`, `version`,
  `description`, `author`. No `agents`/`skills` overrides; default-directory
  discovery finds `.claude/agents/` and `.claude/skills/` directly.
- `.claude/skills/tm-review-changes/SKILL.md` and
  `.claude/skills/tm-review-codebase/SKILL.md` - thin wrapper skills so a
  plugin install can reach the two review workflows, since `workflows/` is
  not an official plugin component type. Each wrapper branches on
  `CLAUDE_PLUGIN_ROOT`: under a plugin it calls the Workflow tool with
  `scriptPath: ${CLAUDE_PLUGIN_ROOT}/workflows/<name>.js` (no `.claude/`
  segment now that the plugin root is `.claude/` itself); otherwise it
  invokes the workflow by name as before.
- `.claude/agents/developer.md` - `description:` frontmatter rephrased to
  remove a colon-space sequence that is invalid YAML
  (`claude plugin validate` failed on it; the runtime loader had tolerated
  it, but a plugin install would silently drop the agent's frontmatter).
- README - third "Getting the team into your repos" option, updated for the
  `.claude/` plugin root and the relaxed tm-install-team handling below.

## Verification performed

- `claude plugin validate .claude` and `claude plugin validate .claude
  --strict`: both pass clean.
- Real, non-interactive install/uninstall cycle in an isolated
  `CLAUDE_CONFIG_DIR` (`claude plugin marketplace add <local path>` +
  `claude plugin install sv-tmueller@claude-template --scope user`, real CLI
  subcommands, not the interactive `/plugin` flow), cleaned up after:
  - `claude plugin details sv-tmueller` reports **Agents (4): reviewer,
    architect, developer, tester** and **Skills (7): tm-advisor,
    tm-grill-me, tm-install-team, tm-kickoff, tm-review-changes,
    tm-review-codebase, tm-to-issues**.
  - The resolved wrapper `scriptPath`
    (`${CLAUDE_PLUGIN_ROOT}/workflows/tm-review-changes.js`) exists on disk
    in the installed plugin cache, confirming the no-`.claude/`-segment path
    is correct under the new root.
- `npm test`: 40/40 passing (workflow helper tests untouched).

## Known limitations

**`tm-install-team` ships but is inert outside a template checkout.**
Default-directory discovery registers all 7 skills; there is no override
mechanism left to exclude one without removing it from `.claude/skills/`,
which the template and config-dir deployment models still need. The skill
is `disable-model-invocation: true` and self-guards on its own precondition
(it checks that `.claude/agents/`, `.claude/skills/`, `.claude/workflows/`,
and `.claude/team-guide.md` exist in the *current working directory* before
doing anything), so under a plugin install in some other project it stops
at that guard and does nothing. Accepted and documented in the README
rather than worked around by moving or duplicating files.

**Wrapper-vs-workflow name collision in the non-plugin models: reasoned,
not empirically proven.** `tm-review-changes` and `tm-review-codebase`
exist as both `.claude/skills/<name>/SKILL.md` (the plugin wrapper) and
auto-discovered `.claude/workflows/<name>.js` commands in the
committed-repo and config-dir deployment roots. I could not exercise this
collision directly: the task constraints permit only `claude plugin ...`
subcommands, and the collision is specific to the non-plugin models, where
the discriminating event (typing `/tm-review-changes` in a live session
with both files present) is an interactive flow outside those subcommands
and outside what I can drive non-interactively here. Reasoning instead of
a live test: the wrapper skill is `disable-model-invocation: true`
(user-invocable only, never auto-triggered by the model) and its
`CLAUDE_PLUGIN_ROOT`-unset branch already delegates to the same workflow by
name (`Workflow tool, name: "tm-review-changes"`), the same one the
auto-discovered command would run. So even in the worst case (both
register and the command listing shows a duplicate entry), invoking either
produces the same workflow run. This does not block shipping; it should be
verified for real on a second, logged-in machine before relying on it
further.

**Real second-machine runtime agent dispatch remains an untested claim.**
`claude plugin details` confirms the manifest registers all 4 agents and 7
skills in an isolated, non-interactive install/uninstall cycle. Whether
`/sv-tmueller:architect`-style dispatch actually invokes the agent in a live,
logged-in session was not (and could not be) verified from this sandbox.

**Deferred path/slash-command references (batch 2, pre-authorized by the
issue's non-goals).** The shared skill and agent bodies are unchanged
beyond the developer.md frontmatter fix and the two wrapper scriptPath/prose
updates above. The ~15 `/tm-X` prose references and `tm-advisor`'s line-86
citation of `.claude/skills/tm-kickoff/SKILL.md` remain as-is: rewriting
them to `/sv-tmueller:tm-X` would break the committed-repo and config-dir
roots, where the bare `/tm-X` form is correct, and no single
deployment-agnostic fix exists. This is the batch-2 path-reference
hardening the issue defers.


sv-tmueller:
## Round 1 review + directed fix (lead, advisor batch #130)

The round-1 implementation is sound but the chosen manifest shape does not
register the agents. The lead spiked the fix on the logged-in machine with the
real `claude plugin` CLI (isolated `CLAUDE_CONFIG_DIR`, cleaned up) and confirmed
the corrected layout. Apply the following on branch `feat/131-package-team-as-plugin`.

### What round 1 got wrong

- `plugin.json` used `source: "./"` (plugin root = repo root) with `agents` and
  `skills` path-override arrays. The `agents` override does NOT register agents:
  `claude plugin details` reports `Agents (0)`. (The dev's own A/B already proved
  this is a real registration gap, not a `details` display bug: the same tool
  counts `Agents (4)` when agents are real files in the plugin's default `agents/`
  directory.)

### Confirmed fix (lead-verified): plugin root = `.claude/`

Spiked layout, installed cleanly, `claude plugin details` reported
`Agents (4)  reviewer, architect, developer, tester` and `Skills (7)`:

- `marketplace.json` stays at repo root `.claude-plugin/marketplace.json`; change
  the plugin `source` from `"./"` to `"./.claude"`.
- Move the plugin manifest to `.claude/.claude-plugin/plugin.json`.
- DROP both `agents` and `skills` overrides. Rely on default-directory discovery:
  `.claude/agents/` (real files -> agents register) and `.claude/skills/` (all
  skills register). Minimal manifest is just `name`/`version`/`description`/`author`.
- User-facing commands are unchanged: `/plugin marketplace add sv-tmueller/claude-template`
  then `/plugin install quinnslab@claude-template`. The repo-root marketplace.json
  redirects the source into `.claude/`.

### Decision A relaxed (in-scope): tm-install-team ships inert

Default `.claude/skills/` discovery registers all 7 skills, including
`tm-install-team`; the override does not exclude it under this layout, and it
cannot be excluded without removing it from `.claude/skills/` (which the template
and config-dir models need). It is `disable-model-invocation: true` and
self-guarded, so it ships present-but-inert. Accept and document this in the PR
known-limitations and the README. Do not move or duplicate files to exclude it.

### Must-fix: developer.md YAML frontmatter is invalid

`.claude/agents/developer.md` `description:` contains `end to end: branch` - a
colon followed by a space, which is invalid YAML. `claude plugin validate .claude`
fails on it ("YAML Parse error: Unexpected token ... loads with empty metadata,
all frontmatter fields silently dropped"). The runtime loader tolerated it so far,
but as a plugin the agent risks loading without its `model`/`tools`/`isolation`/
`skills`. Surgical fix: rephrase to remove the colon-space, e.g.
`end to end (branch, TDD, conventional commits, draft PR).` Then re-run
`claude plugin validate .claude` and confirm it passes clean (catch any other
colon-space frontmatter the same way). This is a frontmatter validity fix, NOT the
deferred path-reference hardening, and it improves all three deployment roots.

### Must-fix: workflow wrapper scriptPath for the new root

Under root = `.claude/`, the workflows are at `${CLAUDE_PLUGIN_ROOT}/workflows/...`
(NOT `${CLAUDE_PLUGIN_ROOT}/.claude/workflows/...`). Update both
`.claude/skills/tm-review-changes/SKILL.md` and
`.claude/skills/tm-review-codebase/SKILL.md`: the `CLAUDE_PLUGIN_ROOT`-set branch
must use `${CLAUDE_PLUGIN_ROOT}/workflows/tm-review-*.js`, and fix the explanatory
prose (it currently says `source: "./"` makes the root the whole repo - now the
root is `.claude/`). Echo and assert the resolved `.js` path exists in the install
test.

### Must-verify: wrapper vs workflow name collision in the non-plugin models

The wrapper skills `tm-review-changes` / `tm-review-codebase` share names with the
auto-discovered `.claude/workflows/*.js` commands. In the committed-repo and
config-dir models both claim `/tm-review-changes` / `/tm-review-codebase`. Verify
empirically what happens (does one shadow the other, do both list, does invoking
the workflow by name still work when a same-named skill exists). If it is benign
(the wrapper delegates to the workflow by name when `CLAUDE_PLUGIN_ROOT` is unset,
so the result is the same), document it. If it breaks or hard-errors, report back
before shipping - do not rename to an ugly `-wf` suffix without flagging it.

### Re-verify (non-interactive, isolated, cleaned up)

In a throwaway `CLAUDE_CONFIG_DIR=$(mktemp -d)`:
`claude plugin marketplace add <local repo path>` ->
`claude plugin install quinnslab@claude-template` ->
`claude plugin details quinnslab`. Assert: `Agents (4)` (architect, developer,
tester, reviewer) and the skills set (7, incl. tm-install-team inert).
`claude plugin validate .claude` clean. `npm test` green. Clean up the config dir.
Use the full binary path `/Users/TM/.local/bin/claude` (the `claude` shell alias is
a stub); never run a bare/interactive `claude` that blocks.

Runtime dispatch of a plugin agent (`/quinnslab:architect`-style) on a real
logged-in second machine remains an honest UNTESTED CLAIM for the human; document
it, do not try to fully prove it here.

sv-tmueller:
## Tester verdict: PASS (round 1)

Independent verification on `feat/131-package-team-as-plugin` @ `7d1e484`:
- `npm test` 40/40; diff touches only 6 files; no `.claude/workflows/` or `*.test.mjs` touched (no test weakened/deleted).
- Isolated `claude plugin` install cycle: `Agents (4)` (architect, developer, reviewer, tester), `Skills (7)`; wrapper `scriptPath` resolves to real `workflows/tm-review-*.js` in the payload; `validate --strict` clean; no contamination of real `~/.claude` / config dirs.
- `developer.md` frontmatter fix confirmed load-bearing (reintroduced the colon-space -> validate fails; shipped fix -> passes). No other agent/skill frontmatter has the same bug.

**Non-blocking finding (1):** `.claude/settings.json` (enables superpowers for the template repo itself) now ships as plugin payload because the root moved to `.claude/`. Empirically inert: after install, the user-scope `enabledPlugins` has only `quinnslab@claude-template`, no superpowers key. Decision D holds, but this isn't in the PR's Known limitations next to tm-install-team.

**Untested-interactively claims (carried to human/batch-2):** (a) wrapper-vs-workflow name collision in the committed-repo/config-dir models (reasoned benign); (b) real runtime dispatch of `/quinnslab:<agent>` on a logged-in second machine; (c) the literal post-merge GitHub marketplace source `/plugin marketplace add sv-tmueller/claude-template` (verified here via local-path source).

sv-tmueller:
Renamed the plugin namespace `quinnslab` -> `sv-tmueller` (commit `83cf9f3`) per the user, to match the GitHub owner. Pure string rename in both manifests, the two wrapper skills, and the README; no structural change. Re-verified the isolated install cycle under the new name: `/plugin install sv-tmueller@claude-template` -> `Agents (4)`, `Skills (7)`, `validate --strict` clean, `npm test` unaffected. Commands now read `/sv-tmueller:tm-advisor` etc. The prior tester PASS and reviewer APPROVE stand (the diff since is namespace strings only).

Issue #262, opened 2026-07-24
Batch: opus-5-model-refresh


sv-tmueller:
Approved contract for this batch, verbatim from the /tm-advisor sign-off. Dispatched 2026-07-24.

**STATUS: run complete. Both packages ready on PR #265. See the final batch report in the comments.**

## Trigger

Opus 5 was released. Question raised: do the agent/skill instructions need updating from Opus 4.8 to Opus 5?

## Findings that shaped the batch

**Nothing is functionally stale.** No agent frontmatter or workflow stage pins a dated model ID. Every seat pins an alias — `model: fable` (architect, reviewer), `model: sonnet` (the five worker seats), and `model: 'fable'` / `'sonnet'` in the three workflow scripts. Aliases resolve to the current model, so no dispatch is running on a superseded one today.

**What is stale is procedure text a human would copy-paste**, in 5 files:

- `.claude/team-guide.md:197` — `/model claude-opus-4-8`, the one literal dated ID in the repo, and the exact string someone types during a Fable outage
- `.claude/team-guide.md:192,195,209` — "Fable costs 2x Opus 4.8", "Fallback: Opus 4.8 at xhigh", "move to Opus 4.8 permanently"
- `.claude/team-guide.md:233` — `claude-sonnet-4-6` as the seatbelt example (Sonnet 5 now exists)
- `.claude/workflows/tm-{map-codebase,review-changes,review-codebase}.js` — header comments citing "the Opus 4.8 fallback"
- `docs/team-guide-rationale.md:71,83` — two `Supports:` quotes that must stay verbatim-matched to the team-guide lines above

The 2x cost claim survives the rename: Fable 5 is $10/$50 per MTok, Opus 5 is $5/$25 — same as Opus 4.8 was. So the sentence stays true with only the model name changed.

**The thing worth challenging.** The model policy puts Fable 5 in the lead seat at 2x price, justified by "affordable only because the lead stays on the bounded tm- machinery," with Opus 4.8 as a documented-but-manual fallback. Opus 5 lands as a step-change over 4.8 on exactly the workload the lead does (long-horizon agentic, planning, code review) at half Fable's per-token cost. That reopens the who-leads question, which a rename would paper over. It is also a policy decision, not an edit — so it is sliced as a recommendation package that changes no policy, and the user decides.

---

### 1. Refresh stale model references in the team guide and workflows

[...]

PR #225, opened 2026-07-06
refactor: trim the injected team-guide to operative rules

sv-tmueller:
Closes #223

## What changed

`.claude/team-guide.md` is injected into every session via `CLAUDE.md`. Its
operative rules were getting diluted by rationale prose and evidence numbers
that belong in a reference doc, not the hot path. This PR relocates every
"why" chunk to a new `docs/team-guide-rationale.md`, leaving team-guide.md as
rules plus pointers (to the rationale doc, or to existing docs: agent files,
`docs/team-architecture.md`, dated research/review docs). No rule is deleted
or weakened; every relocated chunk is copied verbatim into the destination.

## Before / after (byte and token table)

| | bytes | heuristic tokens (bytes/4) |
|---|---|---|
| `.claude/team-guide.md` before | 18,193 | 4,548 |
| `.claude/team-guide.md` after | 12,721 | 3,180 |
| reduction | 5,472 bytes (30.08%) | |
| `docs/team-guide-rationale.md` (new) | 6,934 | 1,734 |

Reduction target was >= 30% (>= 5,458 bytes cut, landing at <= 12,735 bytes).
Landed at 12,721 bytes, 30.08% below the 18,193-byte baseline.

Recomputed injected chain (report-time and contract-time bases converge to
the same total since both offsets are 8,078 bytes against the old
team-guide byte count):

| | bytes | heuristic tokens |
|---|---|---|
| chain total (8,078 fixed + after-bytes) | 20,799 | 5,200 |

## Chunk-to-destination relocation map

| Source section | Chunk | Destination | team-guide.md pointer left behind |
|---|---|---|---|
| Workflow defaults | Permission-mode HTML comment (mode reference table) | `docs/team-guide-rationale.md` "Permission mode reference" | inline `(mode reference: docs/team-guide-rationale.md)` |
| Workflow defaults | Session-hygiene before/after numbers (570,000 / 1,634,150 / 583,848, the 8.2x/2.9x estimate, hit ratios) | `docs/team-guide-rationale.md` "Session hygiene: measured baseline and estimate" | inline `docs/team-guide-rationale.md` |
| Agent team | (no relocation; taxonomy and choreography already live elsewhere) | `.claude/agents/fact-checker.md` (verdict taxonomy), `docs/team-architecture.md` (dispatch choreography) | inline pointers in the fact-checker and perf-investigator bullets |
[...]

Issue #201, opened 2026-07-06
Batch: awesome-agents-adoption


sv-tmueller:
# Batch: awesome-agents-adoption

Approved contract (signed off in chat, 2026-07-06). Implements the four adoption candidates from `docs/reviews/2026-07-05-awesome-claude-agents-survey.md`.

## Refinement decisions (made at sign-off)

- Seat naming: unprefixed role nouns (`docs-writer`, `perf-investigator`), matching the five existing seats. The `tm-` prefix rule covers skills and workflows, not agent seats.
- Health header scope: `tm-review-codebase.js` only. The map workflow forbids severities by design ("a map, not a review"); `tm-review-changes.js` writes no report file and the survey's adaptation rules out schema changes.
- Survey correction: `developer.md` does not have `WebFetch` in its tools list (the survey claimed it did). Package 4 adds the grant.
- Known overlap, accepted: packages 1 and 2 both touch the seat lists in `.claude/team-guide.md`, `README.md`, and `docs/team-architecture.md`. Whichever PR merges second needs a trivial conflict resolution. Rebase nuisance, not a dependency.

### Package 1 - docs-writer seat (size:S)

**Title:** feat: add docs-writer agent seat (survey candidate 1)

**Scope:** Add `.claude/agents/docs-writer.md`, an on-demand execution seat that authors user-facing documentation (READMEs, guides, API docs) from a gap analysis of what is missing, adapted from the source repo's `documentation-specialist`. Dispatched by the lead on demand (for example after a `tm-map-codebase` run), never part of the per-package kickoff pipeline. Update the seat lists in `.claude/team-guide.md` (Agent team + Model policy), `README.md`, and `docs/team-architecture.md`.

**Acceptance criteria:**
- Frontmatter pins `model: sonnet`, `effort: high`; `npm test` (effort-policy test) passes.
- Tools are exactly `Read, Grep, Glob, Write, Edit` - no Bash.
- The agent body defines the job (gap analysis, then author) and a report contract for what it returns to the lead.
- The three seat-list docs name the new seat and its dispatch condition.

**Non-goals:** No new skill or workflow. No auto-dispatch rules. Not added to the kickoff per-package pipeline. No changes to `tm-map-codebase.js`.

### Package 2 - perf-investigator seat (size:M)

**Title:** feat: add perf-investigator agent seat (survey candidate 2)

[...]

PR #214, opened 2026-07-06
research: root-cause the fast session-limit exhaustion on Max 5x

sv-tmueller:
## Summary
- Adds `docs/research/2026-07-06-token-burn-analyze.mjs`, a zero-dependency Node script that streams usage metadata (never message content) from both Claude Code project roots over a closed 14-day window (2026-06-22 to 2026-07-05 inclusive), dedups per `message.id`, and aggregates by project, model, day, cache efficiency, lead-vs-agentType attribution, and per-session peak context. Supports `--json`, `--start`/`--end`, and `--project` for scoped re-runs.
- Adds `docs/research/2026-07-06-token-burn-investigation.md`, the findings report per the sub-plan's 7-section structure, with driver verdicts for all five originally suspected drivers plus one additional finding.
- Headline finding: the dominant driver is unbounded single-session context growth (individual turns up to 1.63M tokens on `lead` sessions), not the five originally suspected drivers - roughly 1.5 to 2.5 orders of magnitude smaller against the two drivers with a direct ratio (driver 1's bootstrap cost, driver 4's injection-chain estimate). Recommendations in section 6 are ranked by estimated savings and traceable to specific findings.

Closes #212

## Test plan
- [x] `npm test` (65/65 pass, exit 0, re-run after the prose fix round)
- [x] `node docs/research/2026-07-06-token-burn-analyze.mjs` runs clean on plain Node, exit 0 (re-run after the prose fix round to confirm no script code changed behavior)
- [x] `node docs/research/2026-07-06-token-burn-analyze.mjs --json` produces valid JSON
- [x] Every quantitative claim in the report maps to a script output section or a recorded one-line command (section 7). Round-1 fact-check found 9 prose-arithmetic mismatches against the (correct) underlying tables; the fix round corrected them but introduced two new sentence-level slips (an over-generalized by-day extreme-days claim and a 1.5-vs-1.41 orders-of-magnitude figure), which round 2 caught and commit 8a2cfd2 corrects with the auditor's own computed values
- [x] Privacy bound verified: no transcript message content read, quoted, or printed at any point (script or manual inspection during development)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

https://claude.ai/code/session_01DzrwcpraNSbcnFo8Fqi74R

PR #265, opened 2026-07-24
docs: refresh stale model references and assess Opus 5 for the judgment seats

### LINKED ISSUE #263: Refresh stale model references in the team guide and workflows
Part of batch #262

## What to build

Every reference to a superseded model in the live team machinery names the current model instead, so the documented fallback procedure and the seatbelt example are correct to copy-paste. The `/model` command a human types during a Fable outage currently names Opus 4.8; the Model policy prose in `.claude/team-guide.md`, the three workflow header comments, and the two `Supports:` quotes in `docs/team-guide-rationale.md` all still describe an Opus 4.8 fallback.

Nothing about which model each seat runs changes. Every seat pins an alias (`fable`, `sonnet`), and aliases already resolve to the current model — this package corrects documentation and procedure text only.

The 2x cost claim in the Model policy needs care: it should be re-verified against current published pricing rather than carried over on assumption. At the time of filing, Fable 5 is $10/$50 per MTok and Opus 5 is $5/$25 (the same as Opus 4.8 was), so the claim holds with only the model name changed — but confirm before rewriting.

## Acceptance criteria

- [ ] `/model claude-opus-4-8` reads `/model claude-opus-5`
- [ ] The three Model-policy sentences naming Opus 4.8 (fallback seat, 2x cost claim, permanent-move trigger) name Opus 5
- [ ] The 2x cost claim is re-verified against current published pricing before it is rewritten; if the ratio no longer holds, the number is corrected rather than the name swapped
- [ ] The `CLAUDE_CODE_SUBAGENT_MODEL` seatbelt example names the current Sonnet ID
- [ ] The three workflow header comments name the Opus 5 fallback
- [ ] Both `Supports:` quotes in `docs/team-guide-rationale.md` match their team-guide sentences verbatim after the edit
- [ ] `grep -rn "opus-4-8\|Opus 4.8\|sonnet-4-6" .claude/` returns nothing
- [ ] `npm test` passes

## Non-goals

- No change to any `model:` frontmatter pin or workflow stage pin
- No change to the Fable->Opus ladder itself, or to which seat runs which model
- No edits to dated records (`docs/reviews/`, `docs/research/`, `docs/plans/`, `docs/architecture/2026-07-05-codebase-map.md`) — those are point-in-time snapshots and stay as written
- No new `opus` entry in the effort-policy test's model map (nothing pins `opus` after this change)

### LINKED ISSUE #264: Recommend whether Opus 5 should take the lead and judgment seats from Fable 5
Part of batch #262

## What to build

A dated comparison under `docs/research/` that answers one question: with Opus 5 released, should the orchestrator, architect, and reviewer seats stay on Fable 5 or move to Opus 5?

The current Model policy puts Fable 5 in the lead seat at 2x the per-token price, justified by "affordable only because the lead stays on the bounded tm- machinery," with Opus 4.8 as a documented-but-manual fallback. Opus 5 is positioned as a step-change over 4.8 on exactly the workload the lead does (long-horizon agentic, planning, code review) at half Fable's per-token cost. That reopens the who-leads question.

Ground the answer in this repo's own measured data (`docs/research/2026-07-06-token-burn-investigation.md` for the aggregate-vs-per-batch token share, `docs/reviews/2026-06-30-orchestration-comparison.md` for the measured orchestration trial) alongside published capability and pricing differences. Cover the second-order cost too: Opus 5 carries documented behavioral shifts — longer default user-facing output, self-verification without being asked, readier subagent delegation, task-scope expansion — that would need prompt adjustments in whichever seats adopt it. A move is not free even where the token math favors it.

End with a single recommendation and a stated confidence level. This package writes a recommendation, not policy: acting on it is a separate decision for the user, in a later batch.

## Acceptance criteria

- [ ] A dated doc exists under `docs/research/` naming the decision, the options, and a single recommendation
- [ ] The cost comparison uses current published per-token pricing for Fable 5 and Opus 5, cited
- [ ] The recommendation is tied to this repo's measured burn data, not to vendor claims alone
- [ ] If adopting Opus 5 in any seat, the doc lists which agent prompts would need adjusting and why
- [ ] The doc states what would have to be true to revisit the decision
- [ ] `.claude/team-guide.md` is unchanged by this package

## Non-goals

- Does not change the Model policy, any frontmatter pin, or any workflow stage
- Does not run a live A/B (no `tm-ab-test` execution, no token spend on a comparison run)
- Does not revisit the Sonnet worker seats or the effort ceiling

sv-tmueller:
Batch #262, run through `/tm-advisor`, plus one follow-up commit requested directly. Two packages, one branch — this session is pinned to a single branch, so both land here as separate commits instead of one PR each. Rationale and cost of that deviation are logged on #262.

Closes #263
Closes #264

## Why

Opus 5 shipped. The question was whether the agent and skill instructions still name the right models. They mostly do — every seat pins an alias (`fable`, `sonnet`), and aliases already resolve to the current model, so nothing was running on a superseded model. What was stale is the procedure text a human copy-pastes, and this batch proved that is not a cosmetic distinction: both judgment-seat dispatches hit `You've reached your Fable 5 limit` mid-run, and the remedy the team guide told the operator to run was `/model claude-opus-4-8`.

## What changed

Seven commits: one for #263, five for #264, one follow-up.

**Package #263 — model references** (`ec14b7a`, 5 files, 10 insertions / 10 deletions)

Eight targeted string edits, in place, with no paragraph re-flow; every replacement is two characters shorter than the original, so the diff stays readable.

- `.claude/team-guide.md` — the `/model` command in the Fable-outage procedure, the fallback-seat line, the 2x cost claim, the permanent-move trigger, and the `CLAUDE_CODE_SUBAGENT_MODEL` seatbelt example
- `.claude/workflows/tm-{map-codebase,review-changes,review-codebase}.js` — the header comment naming the fallback
- `docs/team-guide-rationale.md` — the first `Supports:` quote

**Package #264 — Opus 5 vs Fable 5 for the judgment seats** (`898bcc4`, `d68dedf`, `d1c0301`, `adb8ef5`, `6f73e2d`)

Adds `docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md`, 442 lines. A recommendation only: it changes no policy, no pin, and no workflow stage.

Its recommendation is a **split, not a clean move** — take the lead seat to Opus 5 now, and hold architect, reviewer and the three workflow critic stages on Fable 5 until the effort-policy test gains an `opus` tier. Confidence: medium. The doc also records that the vendor's own model card still ranks Fable 5 above Opus 5 on capability, which is evidence against a full move, labeled as a vendor claim rather than a measurement.

Two substantive corrections it makes to this repo's existing numbers:

- **The repo's published dollar figures overstate Fable's share.** `docs/research/2026-07-06-token-burn-analyze.mjs` prices Opus at $15/$75 and Sonnet at one fifth of Opus; at current pricing Sonnet is three fifths. Re-derived, batch #201's Fable-priced weighted share is **81.9%**, not the published 93.3% — and the uniform-bucket-mix assumption behind that is validated against the source's own numbers (93.1% predicted vs 93.3% measured)
- **Halving that slice cuts batch-level spend by 40.9%** at sticker pricing, or **43.6%** at Sonnet's introductory rate, which is live through 2026-08-31

**Follow-up — the Opus 5 fallback is now usable** (`089f754`, 3 files, 10 insertions / 3 deletions)

The Model policy tells an operator to flip the `fable` pins to `opus` for a longer Fable outage. Following that instruction failed `npm test`: `EFFORT_BY_MODEL` in `effort-policy.test.mjs` had no `opus` key, so any flipped seat tripped `model "opus" has no effort rule`. The documented recovery path could not be taken without also editing the test that polices it — and this batch needed that path twice.

`opus` now carries fable's `xhigh`, so a flipped seat keeps its effort. Verified in both directions: with all five judgment pins flipped to `opus` the suite passes 65/65, and removing the tier reproduces the failure. Both agent frontmatter comments now name Opus 5 explicitly. No existing `effort:` pin was touched.

This also clears the blocker behind #264's split recommendation — see open question 2 below.

## The finding that outlives this PR

**Acting on #264's recommendation is an exception to written Model policy, not a config change.**

An architect arbitration found that commit `7fd326e` (#133), landed one day after the 2026-06-30 Corollary report, moved the lead to Fable with the reasoning that the lead "routes every handoff, refines batches, and makes the dispatch decisions, so it gets the strongest model" — ratifying the Corollary's principle into the policy line "The strongest model in every plan/decision seat." A lead-only move to Opus 5 is an exception to it, and the follow-up amends the orchestrator bullet.

**Merging this PR does not imply that decision.** The doc says so itself.

The recommendation rests on a floor claim, not on cost. The cost argument was conceded outright: the Corollary weighed comparable pressure (1.67x-2.5x against this doc's 2.0x), said "Cost cuts the other way", and rejected the swap anyway.

## Two things a reviewer should know

**The rationale quote was already broken, and `ec14b7a` removes a clause rather than substituting one.** `docs/team-guide-rationale.md:71` claimed to quote the team guide but carried `and weighs correspondingly against Max-plan quota`, a clause the team guide itself lost in `de2d0f4` (#225). The verbatim-match acceptance criterion exposed pre-existing drift, not just a stale model name. It was trimmed to match the team guide rather than restoring the clause upstream, because #225 deliberately shrank that injected file. Decision logged on #262, independently seconded by the reviewer.

**Every remaining "Opus 4.8" string in the repo is a historical record and was deliberately left alone.** A full sweep found 27 hits, all in dated files: measured token-count tables where `claude-opus-4-8` rows are real usage, a price table using the then-current $15/$75 tier, a dated report comparing Sonnet 5 against Opus 4.8, `Co-Authored-By` commit trailers, and the new research doc's deliberate citations. Renaming any of them would put false statements into the record. Nothing in live machinery still names Opus 4.8.

## Open questions for you

1. `.claude/team-guide.md:192` now couples a price ratio to a version name, the same coupling that made this issue necessary. It will go stale at the next model bump. Rewording to the tier ("2x the Opus tier") would decouple it, but criterion 2 mandated naming Opus 5, so widening scope is your call
2. ~~The team guide's other Fable-outage remedy fails `npm test`.~~ **Fixed in `089f754`.** Note this partially undercuts #264's split recommendation, which held architect, reviewer and the critic stages back specifically because of this blocker. The capability argument for holding them (vendor positioning still ranks Fable above Opus 5) is unaffected and stands on its own
3. One residual nit: line 241 of the research doc says "the effort-policy failure" where "blocker" is accurate. Disclosed rather than silently patched

## Verification

- `grep -rn "opus-4-8\|Opus 4.8\|sonnet-4-6" .claude/` returns nothing
- `npm test` 65/65 on the final commit
- Both `Supports:` quotes diffed character by character against their team-guide sentences
- No `model:` frontmatter pin or workflow stage pin changed; no `effort:` pin changed; no dated record edited; every #264 commit touches only the research doc
- **#263:** tester PASS round 1, reviewer APPROVE round 1, zero findings against the diff
- **#264:** tester FAIL -> PASS -> FAIL -> arbitration -> PASS; reviewer CHANGES_REQUESTED -> APPROVE; plus one fact-checker audit (26 claims verified, 3 contradicted) and one architect arbitration
- **`089f754`:** verified by flipping all five judgment pins to `opus` (65/65 pass) and by removing the tier to reproduce the original failure

This repo has no `.github/workflows/`, so there are no CI checks on this PR. `npm test` run locally is the gate.

https://claude.ai/code/session_01QMT3YhDmHtzrnTkfTwVWth

### DIFF
--- a/.claude/agents/architect.md
+++ b/.claude/agents/architect.md
@@ -2,7 +2,8 @@
 name: architect
 description: Advisory lead for approach decisions. Use when an issue needs a sub-plan before implementation, when an issue looks mis-sized or needs splitting (size:L or size:XL), or when developer and reviewer disagree. Read-only; decides approach, never writes code.
 tools: Read, Grep, Glob, Bash
-# Judgment role: Fable 5. Fallback is opus (see team-guide, Model policy).
+# Judgment role: Fable 5. Fallback is opus = Opus 5, same xhigh effort
+# (see team-guide, Model policy).
 model: fable
 effort: xhigh
 ---
--- a/.claude/agents/reviewer.md
+++ b/.claude/agents/reviewer.md
@@ -2,7 +2,8 @@
 name: reviewer
 description: Reviews a work package diff against its issue and sub-plan, in two passes, spec compliance then code quality. Read-only; outputs APPROVE or CHANGES_REQUESTED with numbered file:line findings. Never edits files.
 tools: Read, Grep, Glob, Bash
-# Judgment role: Fable 5. Fallback is opus (see team-guide, Model policy).
+# Judgment role: Fable 5. Fallback is opus = Opus 5, same xhigh effort
+# (see team-guide, Model policy).
 model: fable
 effort: xhigh
 ---
--- a/.claude/team-guide.md
+++ b/.claude/team-guide.md
@@ -189,12 +189,12 @@ Rationale: docs/team-guide-rationale.md.
 
 - Orchestrator (the lead session, including `/tm-advisor` and `/tm-kickoff`):
   Fable 5 (`claude-fable-5`) at xhigh effort. Affordable only because the
-  lead stays on the bounded tm- machinery. Fable costs 2x Opus 4.8 per
+  lead stays on the bounded tm- machinery. Fable costs 2x Opus 5 per
   token. The premium is bounded in aggregate, not per batch
   (docs/research/2026-07-06-token-burn-investigation.md, driver 3).
-- Fallback: Opus 4.8 at xhigh effort (a procedure, not automatic). When
+- Fallback: Opus 5 at xhigh effort (a procedure, not automatic). When
   Fable 5 is unavailable, rate-limited, quota-exhausted, or
-  refuses the workload, switch the lead with `/model claude-opus-4-8`; for a
+  refuses the workload, switch the lead with `/model claude-opus-5`; for a
   longer outage flip the `fable` pins to `opus` in the `architect` and
   `reviewer` frontmatters and the Fable-critic stage of every `tm-` workflow
   (`.claude/workflows/`). Flip back when Fable returns. For a single
@@ -206,7 +206,7 @@ Rationale: docs/team-guide-rationale.md.
 - Cost-based fallback trigger: if Fable 5 stops being included under the
   Max-plan subscription and shifts to metered API billing, do not switch to
   Opus automatically. Measure the lead's actual $/session cost at API rates
-  first, then decide whether to keep Fable or move to Opus 4.8 permanently,
+  first, then decide whether to keep Fable or move to Opus 5 permanently,
   logging the decision and cost here.
 - No session-wide `ultracode` under this policy (a prompt keyword or
   `/effort` menu option, not a slash command). Measured trial:
@@ -230,7 +230,7 @@ Rationale: docs/team-guide-rationale.md.
   never pays Fable rates for worker stages.
 - Do not set `CLAUDE_CODE_SUBAGENT_MODEL`. It flattens every subagent to one
   model, defeating the split above. Use only as a temporary seatbelt (e.g.
-  `claude-sonnet-4-6` before one heavy ad-hoc run); it downgrades the
+  `claude-sonnet-5` before one heavy ad-hoc run); it downgrades the
   architect and reviewer too.
 
 ## How to pick up a task
--- a/.claude/workflows/__tests__/effort-policy.test.mjs
+++ b/.claude/workflows/__tests__/effort-policy.test.mjs
@@ -6,6 +6,11 @@
  * The test reads the real agent frontmatters and workflow sources, so a new
  * agent or workflow stage that omits its pin (and would silently inherit the
  * session effort) fails here instead of shipping.
+ *
+ * `opus` carries fable's xhigh so the documented Fable-outage fallback is
+ * usable: flipping a judgment seat's pin from fable to opus keeps its effort
+ * unchanged and still passes. Without that entry the fallback the Model policy
+ * prescribes fails this suite the moment anyone follows it.
  */
 
 import { test, describe } from 'node:test'
@@ -18,7 +23,7 @@ const __dir = dirname(fileURLToPath(import.meta.url))
 const workflowsDir = join(__dir, '..')
 const agentsDir = join(__dir, '..', '..', 'agents')
 
-const EFFORT_BY_MODEL = { sonnet: 'high', fable: 'xhigh' }
+const EFFORT_BY_MODEL = { sonnet: 'high', fable: 'xhigh', opus: 'xhigh' }
 const WORKFLOW_FILES = ['tm-review-changes.js', 'tm-review-codebase.js', 'tm-map-codebase.js']
 
 // ===========================================================================
--- a/.claude/workflows/tm-map-codebase.js
+++ b/.claude/workflows/tm-map-codebase.js
@@ -16,7 +16,7 @@ export const meta = {
 // areas the scout proposes. There is no per-file fan-out and no loop. Models and
 // effort are pinned per stage, so the session model and effort never leak into
 // the scout or the workers; the single critic runs Fable 5 at xhigh effort (flip
-// its model pin to 'opus' under the Opus 4.8 fallback, see team-guide).
+// its model pin to 'opus' under the Opus 5 fallback, see team-guide).
 //
 // When the repo is too big for MAX_AREAS areas to cover, the leftover paths are
 // reported (coverage.areasDropped, coverage.ceilingReached, and a suggested next
--- a/.claude/workflows/tm-review-changes.js
+++ b/.claude/workflows/tm-review-changes.js
@@ -14,7 +14,7 @@ export const meta = {
 // session-model review produces. Models and effort are pinned per
 // stage, so the session model and effort never leak into the workers; the
 // single critic runs Fable 5 at xhigh effort (flip its model pin to 'opus'
-// under the Opus 4.8 fallback, see team-guide).
+// under the Opus 5 fallback, see team-guide).
 //
 // Invoke with an optional base ref:
 //   Workflow({ name: 'tm-review-changes', args: { base: 'origin/main' } })
--- a/.claude/workflows/tm-review-codebase.js
+++ b/.claude/workflows/tm-review-codebase.js
@@ -17,7 +17,7 @@ export const meta = {
 // fan-out and no loop. Models and effort are pinned per stage, so the
 // session model and effort never leak into the scout or the workers; the
 // single critic runs Fable 5 at xhigh effort (flip its model pin to 'opus'
-// under the Opus 4.8 fallback, see team-guide).
+// under the Opus 5 fallback, see team-guide).
 //
 // When the repo is too big for MAX_AREAS areas to cover, the leftover paths are
 // reported (coverage.ceilingReached, coverage.areasDropped, and a suggested next
--- a/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
+++ b/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
@@ -93,13 +93,17 @@ dispatch, no architect/reviewer): $1.80. Ledger row:
 lines 63-77 weight all four scopes above using historical list-price
 tiers: `OPUS_INPUT = 15`, `OPUS_OUTPUT = 75`, Sonnet `3`/`15`, with Fable
 set to 2x that Opus figure (`30`/`150`). Current published pricing (bundled
-`claude-api` skill, `shared/models.md`, cached 2026-06-24): Opus 5
-`claude-opus-5` $5/$25 per MTok, Fable 5 `claude-fable-5` $10/$50, Opus 4.8
-`claude-opus-4-8` $5/$25 (still active), Sonnet 5 `claude-sonnet-5` $3/$15.
-Cross-checked against this repo's own citation in
-`docs/reviews/2026-06-30-orchestration-comparison.md` ("Corollary": Sonnet
-$3/$15 vs Opus 4.8 $5/$25, checked 2026-06-30) - the two sources agree on
-Sonnet and Opus 4.8.
+`claude-api` skill, `shared/models.md`, which self-describes as cached, not
+live, at line 7, with no embedded cache date): Opus 5 `claude-opus-5` $5/$25
+per MTok, Fable 5 `claude-fable-5` $10/$50, Opus 4.8 `claude-opus-4-8` $5/$25
+(still active). That file carries no Sonnet 5 pricing at all
+(`grep -n '\$' shared/models.md` returns only the Fable 5 and Opus 5 lines);
+Sonnet 5's $3/$15 comes from `shared/model-migration.md:1192` ("unchanged at
+the $3/$15 sticker"), `python/claude-api/README.md:502`, and this repo's own
+`docs/reviews/2026-06-30-orchestration-comparison.md:126-129` ("Corollary":
+Sonnet $3/$15 vs Opus 4.8 $5/$25, checked 2026-06-30). Those three sources
+agree with each other on Sonnet, and the last also agrees with
+`shared/models.md` on Opus 4.8's $5/$25.
 
 The Sonnet leg of the proxy table ($3/$15) already matches current
 pricing exactly, but the Opus/Fable leg does not: at the proxy table,
@@ -129,7 +133,7 @@ were never published for that batch. Under that assumption:
 - Opus 5 is exactly half of Fable 5's price in both directions, so
   replacing every Fable-priced seat with Opus 5 halves that slice's dollar
   contribution. Cutting an ~80-85% slice in half cuts total batch-level
-  spend by roughly 35-45%.
+  spend by roughly 40-42.5%.
 
 That range is the ceiling for option (b) (all three seats moved), at
 batch scope, in dollars, not confirmed Max-plan quota. It reuses
@@ -144,7 +148,7 @@ more than half of nearly every bucket across the whole dataset (a
 different scope, not batch #201's role split), so it is directional
 evidence that lead is plausibly the majority of the Fable-priced group,
 not a number to plug into the formula above. Option (c)'s savings are
-therefore real, bounded above by the same 35-45% ceiling, and smaller if
+therefore real, bounded above by the same 40-42.5% ceiling, and smaller if
 architect and reviewer carry a non-trivial share. Pricing it more
 precisely would need a role-scoped re-run of the token-burn script
 against a fresh batch (section 9).
@@ -152,15 +156,18 @@ against a fresh batch (section 9).
 ## 5. Availability and operating cost
 
 During this batch, two judgment-seat dispatches failed with `You've
-reached your Fable 5 limit`, each forcing a per-call Opus override
-(the Agent tool's `model` param, per the documented fallback procedure).
-Cross-referenced against commit `11003f9` ("docs: document per-call Opus
-override for a single Fable-blocked seat", #228, verified with
-`git show 11003f9`): that commit exists specifically because the same
-workaround already had to be written down once before, in a prior batch.
-Two Fable-limit events across two separate occasions is a recurrence, not
-a first; it is not a measured failure rate, and this document does not
-treat it as one.
+reached your Fable 5 limit`, each forcing a per-call Opus override (the
+Agent tool's `model` param, per the documented fallback procedure). Both
+events, with the exact error string, are recorded on batch issue #262
+alongside the decision to use the per-call override; that issue is the
+auditable record, since this document's sandbox has no GitHub access or
+session transcript to re-derive the claim from. Cross-referenced against
+commit `11003f9` ("docs: document per-call Opus override for a single
+Fable-blocked seat", #228, verified with `git show 11003f9`): that commit
+exists because the same workaround already had to be written down once
+before, in a prior batch. Two events across two separate occasions is a
+recurrence, not a first; it is not a measured failure rate, and this
+document does not treat it as one.
 
 Two standing costs from the pricing source (`claude-api` skill,
 `shared/models.md`):
@@ -177,7 +184,7 @@ Two standing costs from the pricing source (`claude-api` skill,
 
 Opus 5's prompt-cache minimum dropped to 512 tokens (from 1024 on
 Opus 4.8), but this repo's own measured bootstrap contexts run
-12k-22k tokens (drivers 1 and 4 of the token-burn investigation), so that
+12k-22k tokens (driver 1 of the token-burn investigation), so that
 change is marginal here and does not support the recommendation below.
 
 ## 6. Capability
@@ -227,8 +234,10 @@ Opus 5 today (section 2 notes the mid-session #263 rename that makes the
 prose track this). Worth reconciling in the same follow-up that adds the
 `opus` effort-policy entry.
 
-Prompt rework specific to any Opus 5 shift, tied to each documented
-behavior change:
+Prompt rework specific to any Opus 5 shift, tied to each documented behavior
+change (`claude-api` bundled skill, `shared/model-migration.md`, "Migrating
+to Claude Opus 5" > "Behavioral shifts (prompt-tunable)", lines 1029-1098; a
+vendor migration guide, the same weakest evidence class as section 6):
 
 - **Longer default output.** `reviewer.md`'s report contract fixes
   structure (`VERDICT`/`STAGE`/`FINDINGS`) but sets no length bound. A
@@ -285,7 +294,7 @@ review or sub-plan quality the pipeline already guarantees through
 Fable-pinned subagents.
 
 What caps this at medium rather than high: the precise dollar/quota
-savings from moving the lead alone are bounded (up to the 35-45%
+savings from moving the lead alone are bounded (up to the 40-42.5%
 batch-level ceiling in section 4) but not pinned, since driver-3 does not
 publish a lead-only breakdown; "affordable" in the current policy is a
 Max-plan quota claim, and nothing here measures whether Opus 5's lower
@@ -322,9 +331,14 @@ the recommended delegation cap in place.
   `docs/reviews/ab-tests.md`.
 - `docs/reviews/2026-06-30-orchestration-comparison.md` (Corollary,
   addendum, and the $3/$15 vs $5/$25 pricing cross-check).
-- `claude-api` bundled skill, `shared/models.md`, cached 2026-06-24. Not
-  a live query; the file itself recommends the Models API for anything
-  capability-related, which this document did not run.
+- `claude-api` bundled skill: `shared/models.md` (self-described as cached,
+  not live, at line 7, no embedded cache date; recommends the Models API for
+  anything capability-related, which this document did not run);
+  `shared/model-migration.md:1192` (Sonnet 5 pricing) and its "Migrating to
+  Claude Opus 5" > "Behavioral shifts (prompt-tunable)" section, lines
+  1029-1098 (section 7's behavior-shift claims, a vendor migration guide,
+  the same weakest evidence class as section 6's capability claims);
+  `python/claude-api/README.md:502` (Sonnet 5 pricing).
 - `.claude/team-guide.md`, `docs/team-guide-rationale.md`,
   `.claude/agents/architect.md`, `.claude/agents/reviewer.md`,
   `.claude/workflows/tm-review-changes.js`,
--- a/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
+++ b/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
@@ -1,0 +1,339 @@
+# Opus 5 vs Fable 5 for the lead, architect, and reviewer seats - 2026-07-24
+
+A recommendation, not a policy change. Related: issue #264. Nothing here
+edits `.claude/team-guide.md`, any agent frontmatter, or any workflow stage.
+
+## 1. Question and scope
+
+With Opus 5 released, should the orchestrator (lead), architect, and
+reviewer seats stay on Fable 5, or move to Opus 5? These three are the
+Model policy's "judgment" seats: the lead session itself, plus the two
+`fable`-pinned role agents (`architect.md`, `reviewer.md`) and, by
+extension, the three workflow critic stages that also pin `fable`
+(`tm-review-changes.js`, `tm-review-codebase.js`, `tm-map-codebase.js`).
+
+In scope: one recommendation with a stated confidence level, grounded in
+this repo's own measured burn data and current published pricing.
+Out of scope: the Sonnet-pinned worker seats, the `xhigh` effort ceiling,
+any live A/B run (no `tm-ab-test` execution), and any actual edit to the
+Model policy, frontmatter, or workflow scripts. A follow-up issue would
+carry out any change this document recommends.
+
+## 2. What the current policy rests on
+
+`.claude/team-guide.md`, Model policy, as of this session:
+
+> Fable 5 (`claude-fable-5`) at xhigh effort. Affordable only because the
+> lead stays on the bounded tm- machinery. Fable costs 2x Opus 5 per
+> token. The premium is bounded in aggregate, not per batch
+> (docs/research/2026-07-06-token-burn-investigation.md, driver 3).
+
+Note on that quote: issue #263 landed on this same branch while this
+document was being written and renamed the fallback model throughout
+`team-guide.md` from Opus 4.8 to Opus 5 (a stale comparison target, not a
+dollar-math error). The line above already reflects that edit; earlier
+drafts said "2x Opus 4.8 per token." Opus 4.8 and Opus 5 are priced
+identically ($5/$25 per MTok, section 4), so the "2x" figure is unchanged
+by the rename.
+
+Two things to hold apart when reading "affordable":
+
+- It is a **Max-plan subscription quota claim**, not a dollar claim. The
+  policy's own "Cost-based fallback trigger" bullet says as much: if
+  Fable 5 ever moves off the Max-plan subscription to metered API
+  billing, "measure the lead's actual $/session cost at API rates first,"
+  rather than assuming the quota reasoning still applies.
+- "Bounded in aggregate, not per batch" is a scope claim, backed by one
+  finding (section 4): Fable's token share is small across 14 days and 25
+  projects, but not small inside a single kickoff batch.
+
+## 3. Options
+
+Three concrete seat assignments, not general positions.
+
+- **(a) Stay on Fable everywhere.** Lead, architect, reviewer, and all
+  three workflow critics stay `fable`. The status quo.
+- **(b) Move all three seats to Opus 5.** Lead switches its session model;
+  `architect.md`/`reviewer.md` frontmatter and the three critic stages
+  flip `model: fable` to `model: opus`.
+- **(c) Split: lead moves to Opus 5, architect/reviewer/critics stay on
+  Fable 5.** Priced in section 4, justified in section 8.
+
+A fourth combination (architect/reviewer to Opus 5, lead stays Fable) was
+rejected as a candidate: section 7 shows the frontmatter/critic-stage move
+is the one blocked by today's failing effort-policy test, while the
+lead's own model is not test-pinned at all. Leading with the change that
+needs no code fix first is the lower-risk split.
+
+## 4. Cost, from this repo's numbers
+
+Three scopes, in order of how much of the pipeline they cover.
+
+**14-day, 25-project aggregate.** From
+`docs/research/2026-07-06-token-burn-investigation.md` section 4/5:
+Fable's raw token share is 5.5%, its weighted-proxy-dollar share is 18.7%.
+Lead attribution alone (section 3) accounts for 1,105,250 input /
+9,757,736 output / 41,496,802 cache_creation / 1,156,442,742 cache_read
+across 4,535 deduped lines, "more than half of all buckets except
+cache_creation" - the single largest attribution bucket in the dataset.
+
+**Single batch (#201).** Same doc, section 5, driver 3: at batch scope,
+Fable-priced roles (lead + architect + reviewer) flip to 57.5% of raw
+tokens and 93.3% of weighted-proxy dollars, against Sonnet-priced roles
+(developer + tester) at 42.5% raw / 6.7% weighted. This is the finding the
+"bounded in aggregate, not per batch" policy line already cites.
+
+**Single package.** `docs/reviews/2026-07-13-ab-plan-status-parser.md`,
+Arm A (full kickoff pipeline): weighted cost $7.55, of which $6.44 is
+Fable-priced (85.3%) and $1.11 Sonnet-priced. Arm B (developer-only
+dispatch, no architect/reviewer): $1.80. Ledger row:
+`docs/reviews/ab-tests.md`.
+
+**Mandatory price correction.** `docs/research/2026-07-06-token-burn-analyze.mjs`
+lines 63-77 weight all four scopes above using historical list-price
+tiers: `OPUS_INPUT = 15`, `OPUS_OUTPUT = 75`, Sonnet `3`/`15`, with Fable
+set to 2x that Opus figure (`30`/`150`). Current published pricing (bundled
+`claude-api` skill, `shared/models.md`, cached 2026-06-24): Opus 5
+`claude-opus-5` $5/$25 per MTok, Fable 5 `claude-fable-5` $10/$50, Opus 4.8
+`claude-opus-4-8` $5/$25 (still active), Sonnet 5 `claude-sonnet-5` $3/$15.
+Cross-checked against this repo's own citation in
+`docs/reviews/2026-06-30-orchestration-comparison.md` ("Corollary": Sonnet
+$3/$15 vs Opus 4.8 $5/$25, checked 2026-06-30) - the two sources agree on
+Sonnet and Opus 4.8.
+
+The Sonnet leg of the proxy table ($3/$15) already matches current
+pricing exactly, but the Opus/Fable leg does not: at the proxy table,
+Opus is 5x Sonnet's input price ($15 vs $3); at current prices, Opus is
+1.67x ($5 vs $3), i.e. Sonnet is 1/5 of Opus in the proxy and 3/5 of Opus
+now. **Every dollar figure quoted above (18.7%, 93.3%, $7.55, $6.44)
+overstates the Fable/Opus share relative to current pricing.** Naively
+rescaling one of those figures (dividing $6.44 by some factor) is not
+done here: the single-package report only publishes already-weighted
+dollars, not the raw per-model token counts a correct re-derivation
+needs. Batch #201's driver-3 finding is different: it publishes the raw
+split itself (57.5%/42.5%), so it can be re-derived.
+
+**The counterfactual, worked from batch #201's raw split.** Assumption
+named up front: a uniform bucket mix across role classes (input, output,
+cache_creation, cache_read tokens in the same relative proportion for
+Fable-priced and Sonnet-priced roles), since per-role, per-bucket splits
+were never published for that batch. Under that assumption:
+
+- Raw split: 57.5% Fable-priced tokens, 42.5% Sonnet-priced.
+- Fable is 3.33x Sonnet at current prices ($10/$50 vs $3/$15, both
+  directions), against the proxy table's implied 10x.
+- Fable-priced share of weighted spend at current prices:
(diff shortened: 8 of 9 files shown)


sv-tmueller:
## Tester verdict, round 1 — issue #263: PASS

Commit under test: `ec14b7a1d16fedff642e1cd6de26bcb73ccc8cc9`. Read-only verification on a detached checkout of that exact SHA.

**Findings: none** (must-fix or otherwise).

### What was reproduced

| # | Criterion | Result |
|---|---|---|
| 1 | `/model claude-opus-4-8` -> `/model claude-opus-5` | confirmed via `git show -- .claude/team-guide.md` |
| 2 | Three Model-policy sentences name Opus 5 | confirmed (fallback seat, 2x cost claim, permanent-move trigger) |
| 3 | 2x cost claim re-verified, not assumed | independently read from the bundled `claude-api` skill: Fable 5 `$10/$50 per MTok`, Opus 5 `$5/$25 per MTok`. 10/5 = 2, 50/25 = 2 |
| 4 | Seatbelt example names current Sonnet | now `claude-sonnet-5` |
| 5 | Three workflow header comments | all read "under the Opus 5 fallback, see team-guide" |
| 6 | Both `Supports:` quotes verbatim | first now matches team-guide verbatim; second byte-identical to before and still matching |
| 7 | `grep -rn "opus-4-8\|Opus 4.8\|sonnet-4-6" .claude/` | no output, exit 1 |
| 8 | `npm test` | 65/65 pass, exit 0 |

### Non-goals held

`.claude/agents/` untouched. All workflow stage pins remain `'sonnet'` / `'fable'` — no `'opus'` introduced. `effort-policy.test.mjs` and `helpers.test.mjs` unchanged. Pre-existing quote drift at `docs/team-guide-rationale.md` lines 91, 104-106, 127-128 confirmed left alone.

Diff is 5 files, 10 insertions / 10 deletions — one changed line per substitution site, no paragraph re-flow (verified with `git diff -U0`).

Every version-less `Opus` mention in the Model policy was correctly left unversioned: the per-call override, "the ladder is Fable -> Opus", "if Opus is also down", "preferring an Opus lead", and the `opus` alias references.

### Nit, not a defect in this PR

The whole-repo sweep found **five** files still carrying stale references, not the four the issue enumerated. The fifth is `docs/research/2026-07-06-token-burn-analyze.mjs`, dating to commit `41a3d90` (#214). It sits under `docs/research/`, which the non-goals explicitly exempt, so leaving it alone is correct. The gap is in the acceptance criteria I wrote, not in the work — recording it so the list is right if anyone reuses it.

### Untested claims

None. All eight criteria and all listed non-goals were reproduced with concrete commands.

---

sv-tmueller:
## Reviewer verdict, round 1 — issue #263: APPROVE

Scoped to commit `ec14b7a` only; the #264 research file on the shared branch was excluded and never inspected.

### Pass 1 — spec compliance: clean

All eight criteria independently reproduced. No scope creep, no under-delivery. Every line the sub-plan marked "must not change" (team-guide 203, 208, 215, 216-217) is intact and still correct, since each uses the unversioned name or the alias.

Two checks worth surfacing:

- Criterion 4 corroborated in-repo, not just by assertion: `docs/team-guide-rationale.md:110` states "The `sonnet` alias resolves to Sonnet 5", consistent with the `claude-sonnet-5` ID form now used.
- Criterion 8's safety is structural, not incidental: the effort-policy parser at `effort-policy.test.mjs:76` selects only lines containing both `label:` and `model:`, and the edited comment lines contain neither. `EFFORT_BY_MODEL` is untouched.

A broader sweep than the criterion required (`claude-(opus|sonnet|haiku|fable)-[0-9]` plus `(opus|sonnet|haiku) ?4` across `.claude/`, `CLAUDE.md`, `AGENTS.md`, `README.md`, and the rationale) returns only the three current IDs.

### On the arbitration call

The reviewer was asked to second-guess the lead's decision to trim the rationale quote rather than restore the dropped clause upstream, and **agrees with trimming**. `de2d0f4` confirms the history: it rewrote the team-guide bullet and dropped `and weighs correspondingly against Max-plan quota` while leaving the rationale quoting the old text. Restoring it would re-inflate the file #225 deliberately shrank and would exceed the signed-off three sentences. The chunk is not weakened — its body still supports the surviving claim with the raw/weighted split, and the quota reasoning lives one heading down under "Cost-based fallback trigger: why it is quota, not dollars".

### Pass 2 — quality

Model policy reads coherently end to end after five substitutions. No em dashes, no cliche phrases, no trailing whitespace introduced. Commit message is correct Conventional Commits.

### Findings

**1. `.claude/team-guide.md:192` — should-fix.** The rewritten claim cites `docs/research/2026-07-06-token-burn-investigation.md` driver 3, but that source computes the ratio against Opus 4.8 pricing and literally quotes "2x Opus 4.8". Both statements hold only if Opus 5 is priced identically to Opus 4.8. Criterion 3 required re-verifying this and the tester reports it verified, but no trace of the check exists in the artifact. **Addressed:** provenance recorded in the PR body (source file and figures). The research doc is a dated record and stays untouched.

**2. commit `ec14b7a` message body — should-fix.** Silent on the one edit that removes content rather than substituting it: dropping the quota clause from the rationale quote. That was an arbitration decision logged on #262, so repo history carries no why. **Addressed:** stated explicitly in the PR body rather than amending a pushed commit.

**3. `.claude/team-guide.md:192` — nit, for the human reviewer.** The claim couples a price ratio to a version name, which is exactly the coupling that made this issue necessary; it will go stale again at the next model bump. Rewording to the tier ("2x the Opus tier") would decouple it, but acceptance criterion 2 mandates naming Opus 5, so it is out of scope here. **Not actioned** — genuinely worth considering, and it is your call whether to widen the scope.

**4. Issue #263 criterion 7 — nit.** The grep is scoped to `.claude/`, so it cannot catch the in-scope prose file `docs/team-guide-rationale.md`. It passed only because criterion 6 covered that file separately. A gap in the criteria I wrote, not in the work.

Findings 3 and 4 are both about the acceptance criteria rather than the diff. Together with the tester's nit about the under-enumerated file list, that is three separate defects in how this package was specified and zero in how it was built.

---

sv-tmueller:
## Tester verdict, round 1 — issue #264: FAIL

Commit under test: `898bcc4e726f1b6704336bb40f1f594789c48dda`. `npm test` 65/65, no regression. Criterion 6 satisfied — `git show --name-only 898bcc4` lists only the new research doc; team-guide.md appears in `git diff origin/main...HEAD` solely via #263's `ec14b7a`.

### Must-fix

**1. Arithmetic error in the batch-level savings estimate** (section 4, repeated at lines 132, 147, 288-289).

The doc derives a Fable-priced weighted share of "roughly 80-85%" and claims that halving it "cuts total batch-level spend by roughly 35-45%". Savings from halving a share S is S/2, so for S in [80%, 85%] the correct range is **40-42.5%**, not 35-45%.

```
python3 -c "print(0.80/2, 0.8185/2, 0.85/2)"
0.4 0.40925 0.425
```

The tester's own recomputation of the share itself checks out: `(0.575 x 3.333)/(0.575 x 3.333 + 0.425) = 0.8185`, so "roughly 80-85%" is a fair rounding of that point estimate. The error is only in the second step. It propagates into the recommendation's confidence statement ("bounded above by the same 35-45% ceiling"), so it is load-bearing, not cosmetic.

**2. Uncited source for section 7's behavioral-shift claims.**

"Longer default output", "self-verification without being asked", "readier delegation", and "task-scope expansion" all come from `shared/model-migration.md`'s "Migrating to Claude Opus 5" section (verified at lines 1031, 1064-1074), but that file is never named in the doc, including in the section 10 Sources list.

```
grep -n model-migration docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
(no matches)
```

Every other claim in the doc is pinned to file and line, so this breaks its own citation discipline and leaves acceptance criterion 4's "why" unauditable.

### Should-fix

**3. Sonnet 5 pricing attributed to a source that does not contain it.** Section 4 attributes `$3/$15` to `shared/models.md`, but `grep -n '\$3\|\$15\b'` on that file returns nothing — it has no Sonnet 5 pricing. The figure is real and correctly sourced from `docs/reviews/2026-06-30-orchestration-comparison.md`, but the doc's claim that "the two sources agree on Sonnet and Opus 4.8" overstates verification for Sonnet, where only one of the two named sources carries the number.

### Nits

4. 339 lines against a 200-300 target, partly from repeating the incorrect figure three times.
5. "drivers 1 and 4" cited together for the 12k-22k bootstrap range; that range is entirely driver 1. Driver 4's own figure is ~6,220 tokens, a different and smaller number.

### What passed

- Proxy-price constants confirmed: `OPUS_INPUT = 15`, `OPUS_OUTPUT = 75`, Sonnet 3/15, Fable = Opus x 2
- Batch #201 figures (57.5%/42.5% raw, 93.3% weighted) and the 14-day aggregate (5.5%/18.7%) trace verbatim to their cited sections
- Uniform-bucket-mix assumption named explicitly, not buried
- Effort-policy blocker confirmed, including the load-bearing distinction: the test constrains frontmatter and stage pins but has no test referencing the lead at all, so the recommendation's split holds
- `opus` alias claim is properly hedged as "not independently confirmed", matching the sub-plan's flag
- Commit `11003f9` characterisation accurate
- Two Fable-limit events explicitly not converted into a failure rate
- Vendor capability claims labeled as the weakest evidence class
- No scope leakage; consistently phrased as recommendation, not enactment

### The highest-consequence check passed

The self-verification distinction was handled correctly. The vendor guide makes a broad claim ("delete your verification scaffolding... the same applies to harness-level scaffolding"). The doc correctly separates that from the workflow critic's instruction at `tm-review-changes.js:145`, which verifies the parallel Sonnet reviewers' findings rather than the critic's own output, and explicitly says "Do not remove it." Acting on the vendor guidance without that distinction would have damaged the review pipeline.

### Untested claims

- That two dispatches hit the Fable limit this session — no transcript access. The doc's treatment (not converting to a rate) is correct regardless of the exact count. **Lead note: I can confirm both events; they are recorded on #262.**
- Whether Opus 5's lower per-token price translates to lower Max-plan quota consumption — flagged in the doc as unmeasured, not verifiable from the worktree
- Any capability comparison on this repo's own judgment tasks — the doc states no such comparison exists and relies on vendor positioning, an acknowledged gap rather than a hidden claim

A fact-checker pass is running in parallel. Findings from both will go into a single fix round.

---

sv-tmueller:
## Fact-checker verdict — issue #264: UNGROUNDED

Dispatched alongside the tester because this package's entire value is claim accuracy. Independent audit, same commit `898bcc4`. **26 VERIFIED, 3 CONTRADICTED, 3 UNVERIFIED, 1 correctly labeled as hedged.**

The three contradictions were derived independently and land on the same defects the tester found, including the identical corrected range. That convergence is the useful signal here.

### CONTRADICTED

**1. Sonnet 5 pricing attributed to a file that does not contain it.**

```
grep -n '\$' shared/models.md
```
Returns only line 71 (Fable $10/$50) and line 73 (Opus 5 $5/$25). No Sonnet 5 price anywhere in that file. The `$3/$15` figure is correct, but it lives in `shared/model-migration.md:1192` and `python/claude-api/README.md:502`, and is corroborated by this repo's own `docs/reviews/2026-06-30-orchestration-comparison.md:126-129`.

**2. The 35-45% savings range does not follow from the doc's own inputs.**

Recomputed exactly as the doc instructs: `S = (0.575 x 3.333)/(0.575 x 3.333 + 0.425) = 0.8185`, inside the doc's own "roughly 80-85%" bracket — that part is right. But cutting an S slice in half reduces total spend by S/2: 40% at S=0.80, 42.5% at S=0.85, 40.9% at the doc's own point estimate. **The 35% floor would require S≈70% and the 45% ceiling S≈90%; neither is reachable from the doc's own bracket.**

**3. The 12k-22k bootstrap range mis-cites driver 4.** Driver 1's per-dispatch range (architect 12,308-12,416; developer 21,357-22,314; tester 17,707-17,973; reviewer 12,357-12,664) matches 12k-22k exactly. Driver 4's own figures are a median lead-session first-request context of 42,947 tokens and a median subagent first-request context of 16,503 — the lead figure is roughly double the claimed upper bound. The paragraph's conclusion is unaffected (both dwarf the 512-token threshold under discussion), but the attribution is wrong.

### UNVERIFIED

- **The two Fable-limit events.** `gh` is not installed in the agent sandbox and no session transcript is readable. **Lead resolution: I can confirm both events directly — they are recorded on #262 with the exact error string. The fix round will cite that issue as the record so the claim becomes auditable.**
- **`Related: issue #264`** — same tooling limitation.
- **The "cached 2026-06-24" date.** `shared/models.md` self-describes as cached (line 7) but carries no embedded date; the date appears in the skill's own header, not the file the doc cites. The "cached, not live" characterisation is sound; the date attribution is not.

### Notable VERIFIED items

Worth recording because they are the load-bearing ones:

- Every in-repo measured figure traces verbatim: batch #201's 57.5%/93.3% and 42.5%/6.7%; the 14-day 5.5%/18.7%; the lead's per-bucket attribution across 4,535 deduped lines; Arm A $7.55 ($6.44 Fable / $1.11 Sonnet) and Arm B $1.80; the 297-agent, ~5.64M-token spend-cap failure
- The proxy-price constants are exactly as claimed: `OPUS_INPUT = 15`, `OPUS_OUTPUT = 75`, Sonnet 3/15, Fable at 2x Opus
- **The effort-policy distinction that carries the recommendation holds.** The test reads only `.claude/agents/*.md` frontmatter and the three workflow files, with no reference anywhere to a lead session or `/model` setting. Substituting `model="opus"` reproduces the quoted failure message
- The `opus` alias claim is correctly hedged rather than asserted — and separately, `shared/models.md:120` does map bare `opus` to `claude-opus-5`, distinct from the `opus 4.8` row at 122
- Commit `11003f9`'s own body confirms it was written in response to a real prior mid-batch Fable-quota block, supporting the doc's "recurrence, not a first" framing
- The uniform-bucket-mix assumption is the first sentence of its paragraph, not buried

Dispatching one consolidated fix round covering both audits.

---

sv-tmueller:
## Tester verdict, round 2 — issue #264: PASS

Commit under test: `d68dedf9b7d54f517ec5e7810572c065640d16a4`. `npm test` 65/65. All six acceptance criteria satisfied.

### Round 1 findings — all seven resolved

| # | Finding | Status |
|---|---|---|
| 1 | Savings arithmetic | **Fixed.** `40-42.5%` at lines 136, 151, 297; `grep -n "35-45"` returns nothing. Recomputed independently: `(0.575 x 3.33)/(0.575 x 3.33 + 0.425) = 0.8185`, halved = 40.0-42.5%. No dangling reasoning tied to the old range's width |
| 2 | Uncited `model-migration.md` | **Fixed.** Cited inline at lines 237-240 and in Sources at 337-340, verified against the real file (`## Migrating to Claude Opus 5` at line 902, `### Behavioral shifts` at 1029). Labeled as a vendor migration guide, consistent with section 6 |
| 3 | Sonnet 5 pricing attribution | **Fixed.** All three real sources verified directly. The overstated "two sources agree" is replaced with the accurate narrower claim |
| 4 | Driver 4 citation | **Fixed.** Line 187 cites driver 1 only, verified against `token-burn-investigation.md:273` |
| 5 | Cache-date attribution | **Fixed.** The unpinnable date is removed; the cached-not-live caveat survives and matches `shared/models.md:7` |
| 6 | Fable-limit events auditable | **Fixed.** Cites #262 as the record, still explicitly declines to treat two events as a failure rate |
| 7 | Length | **Judged acceptable.** +14 net lines, all traceable to citation expansion for findings 2, 3, 5 and 6. No unrelated padding in the diff |

### Regression check — clear

The fix touched 39 lines and removed 25 without damaging anything that passed round 1:

- Recommendation and medium confidence unchanged
- **The self-verification distinction is intact** — section 7's bullet is untouched by the diff, and `tm-review-changes.js:145` still reads "verify it against the actual diff" for the parallel reviewers' findings. This was the highest-consequence content in the doc and the fix round left it alone, correctly
- Every previously verified figure independently re-traced and matching: 57.5%/42.5% raw, 93.3% weighted, 5.5%/18.7%, the lead's per-bucket attribution, Arm A $7.55 ($6.44/$1.11), Arm B $1.80, the 297-agent spend-cap failure, and the proxy constants
- Uniform-bucket-mix assumption still named up front; `opus` alias claim still hedged; no scope leakage

Criterion 6 confirmed the strict way: `git show --name-only` on both `d68dedf` and `898bcc4` lists only the research doc. `.claude/team-guide.md` appears in neither.

### Incidental resolution

Round 1's two audits disagreed about what driver 4 contains. The re-test finds it is a distinct ~6,220-token estimate that never belonged in that paragraph. Since the fix dropped the citation entirely, the disagreement is moot — but the safe resolution turned out to be the correct one.

### Nit, not blocking

Line 100 says `grep -n '\$' shared/models.md` "returns only the Fable 5 and Opus 5 lines". It actually returns three lines; the third is an unrelated curl example containing `$ANTHROPIC_API_KEY`. The substance is right — that file has no Sonnet pricing — but "only" overstates the literal output.

### Untested claims

- **The #262 citation.** The agent sandbox has no GitHub access, so it could not confirm that #262 records the two Fable-limit events. **Lead resolution: I can confirm it directly — both events are recorded on #262 with the exact error string, along with the decision to use a per-call Opus override.** The agent cross-checked what it could, verifying `11003f9` exists and matches the doc's description of it
- Section 6's capability claims remain vendor-positioning only, which the doc states outright as its weakest evidence class rather than concealing

---

sv-tmueller:
## Reviewer verdict, round 1 — issue #264: CHANGES_REQUESTED

Scoped to `898bcc4` and `d68dedf`. Pass 1 spec compliance clean — all six criteria met, non-goals hold by construction (`git diff ec14b7a d68dedf --stat` touches one file), and the doc recommends rather than enacts. `npm test` 65/65.

Because `d68dedf` reached the branch without a developer self-verification report, the reviewer re-checked all seven round-1 fixes from source rather than trusting the round-2 report. All confirmed.

### Must-fix

**1. Line 208 — the "full inventory" grep excluded the repo root.** Section 7 claims a full inventory via `grep -rn "fable\|Fable" .claude/ docs/`. That scope misses `README.md`, which carries seven Fable references including line 77 (`L["Lead - main session<br/>fable, xhigh..."]`), and `docs/team-architecture.md:33` (`LEAD - main session (fable, xhigh effort)`). **Both hardcode the lead's model — the exact seat the doc recommends moving now.** A maintainer working the follow-up from this inventory would ship a repo whose user-facing README states the lead runs Fable, against CLAUDE.md's rule that code and doc are reconciled in the same PR.

### Should-fix

**2. Lines 108-109 — Sonnet's introductory rate is live right now.** The doc says the proxy table's Sonnet leg "already matches current pricing exactly", but its own cited source (`shared/model-migration.md:1192`) reads "unchanged at the $3/$15 sticker (introductory $2/$10 per MTok applies through 2026-08-31)". Today is inside that window, and only Sonnet has an intro rate. At $2/$10, Fable is 5x Sonnet, the Fable-priced weighted share is 87.1%, and the saving is **43.6% — above the 42.5% the doc calls a ceiling** at line 138.

**3. Lines 120-136 — the uniform-mix assumption is validatable, and it passes.** Under uniform mix at proxy prices the predicted Fable-priced weighted share is `(0.575 x 10)/(0.575 x 10 + 0.425)` = 93.1%, against the 93.3% the investigation measured. A 0.2-point match validates the assumption to within rounding. The doc instead hedges and widens its range on an assumption it could have retired.

**4. Lines 288-294 — the Corollary is cited for its premise while its conclusion cuts the other way.** `docs/reviews/2026-06-30-orchestration-comparison.md:112` is titled "Corollary: the same argument favors keeping Opus as lead", and it uses the very residual this doc names (the lead's un-delegated judgment) to argue for keeping the *stronger* model in the lead seat. Borrowing half the argument to reach the opposite conclusion without saying so weakens an otherwise honest doc.

**5. Lines 296-304 — the confidence-cap list omits the largest uncertainty.** No first-party capability comparison exists and vendor positioning ranks Fable above Opus 5, but section 8 uses that only to justify holding architect and reviewer, never as a cap on the seat it does recommend moving. The cached, undated pricing source is missing from the list too.

**6. Lines 1-5 — no summary at the top.** A human deciding this reads 275 lines before reaching the recommendation. Two decision-critical facts are never stated plainly anywhere: the lead move is a one-command, zero-code change, and it is instantly reversible — which is the strongest argument for acting under unmeasured capability risk.

**7. Lines 258-262 — the 297-agent figure is double-counted.** 297 agents / 5.64M tokens / ~51 minutes are the *combined* totals for both arms; the source says Arm A accounted for 9 of those 297, so the unbounded arm was 288. Attributing all 297 to the unbounded construction while contrasting it against "the bounded arm's 9" counts Arm A twice.

**8. Lines 276-294 — dollar-denominated rationale under a Max plan.** The recommendation's cost argument is in dollars, but this repo runs on a subscription where, by the doc's own admission at lines 299-301, nothing measures whether a lower price buys more quota. The argument that actually bites operationally — Opus 5's separate rate-limit bucket against two Fable-limit events — sits in a supporting section rather than the recommendation's reasoning.

### Nits

9. Line 100 — the grep returns three lines, not two (confirmed from round 2). Substance correct.
10. Line 228 — the alias table is cited as lines 116-134; it runs 114-138. The load-bearing row is inside the cited range, so the claim holds.
11. Lines 258-259 — "the one measured over-spawn failure in this repo's history" overstates the source, which says it measures the construction's behavior on Sonnet 5, not a live session's free-form choice.

### On the three judgment questions put to the reviewer

- **The split is evidence-driven, not a compromise.** The two halves rest on different verified facts: a real test blocker for architect/reviewer/critics, and containment plus the lead's measured token dominance for the lead
- **Section 6 engages the vendor's Fable-over-Opus-5 ranking honestly** and states outright that it argues against the move
- **Medium confidence is defensible, arguably understated** — the lead move is one command, zero code, instantly reversible, a fact the doc never states plainly (hence finding 6)

Fix round dispatched covering findings 1-8.

---

sv-tmueller:
## Tester verdict, round 3 — issue #264: FAIL

Commit under test: `d1c0301`. `npm test` 65/65. Seven of the eight reviewer findings fixed and independently verified; one is partially fixed in a way that makes the doc worse rather than better.

### Reviewer findings 1-8

| # | Status |
|---|---|
| 1 | **Fixed.** Repo-root grep re-run independently; `README.md` (7 hits, lead node at line 77) and `docs/team-architecture.md:33` now listed. A full scan of `AGENTS.md`, `CLAUDE.md`, `NEW-PROJECT-SETUP.md`, `package.json` and `.claude-plugin/` found zero further lead-seat hardcodes |
| 2 | **Fixed, math verified.** At sticker: share 81.85% -> saving 40.9%. At intro $2/$10: share 87.12% -> saving 43.6%. Both match the doc exactly. 42.5% is no longer mislabeled a ceiling |
| 3 | **Fixed, math verified.** 93.117% predicted against 93.3% measured. The hedge is replaced by the point estimate, not merely supplemented |
| 4 | **Partially fixed — see must-fix below** |
| 5 | **Fixed.** Both missing caps present at lines 355-368 |
| 6 | **Fixed.** Four-line summary at lines 3-10: recommendation, medium confidence, the blocker, and reversibility |
| 7 | **Fixed** for the agent count (288 of a combined 297); see nit for the token figure |
| 8 | **Fixed.** Max-plan caveat and the availability argument both now in section 8's own reasoning |

Nits 9-11 all confirmed fixed against the actual files.

### Must-fix

**The Corollary rebuttal rests on a false premise** (lines 336-353).

The added paragraph's central distinguishing claim is that "that report's swap had no offsetting pressure". The source says otherwise. `docs/reviews/2026-06-30-orchestration-comparison.md:126-133`:

> Cost cuts the other way. Sonnet 5 is cheaper per token than Opus 4.8 on the API: $3/$15 ... $2/$10 introductory ... against Opus 4.8's $5/$25 ... Net: the judgment-escalation argument favors keeping Opus 4.8 as the default lead

So the source **documents cost pressure of comparable magnitude (1.67-2.5x against this doc's 2x), weighs it explicitly, and still rejects the swap.** Asserting that pressure was absent misrepresents the very source being distinguished, in the section where the recommendation confronts its strongest counter-argument.

The second distinguishing point (vendor framing of Opus 5 as narrowing the gap) is fair but is marketing language, not a capability delta.

The tester's judgment, which I asked for explicitly: this reads as a rationalisation built on an inaccurate premise rather than an honest engagement.

### Nit

The same sentence that correctly fixed the agent count attaches "about 5.64M subagent tokens" to "the unbounded arm alone" without the "of a combined" qualifier. The source only ever publishes 5.64M as a combined total. Same over-attribution pattern finding 7 was raised to fix, applied to a different number.

### Regressions — none

Recommendation and confidence unchanged. Self-verification distinction intact. Every previously verified figure re-traced to source. `opus` alias claim still hedged. Criterion 6 confirmed across all three commits.

### On length — an explicit opinion, as requested

417 lines against a 200-300 target, up from 353. Most of the +64 is load-bearing: the summary, the inventory correction, the cross-check, the confidence caps, the Max-plan paragraph. The 40.9%/43.6% pair is restated three times where two would do, worth about 4-6 lines. But the doc has drifted further past its target every round (300 -> 353 -> 417), and the next round should set a harder line. The Corollary paragraph should be **fixed rather than shortened** — trimming inaccurate content does not help.

---

**Routing note.** Rather than send a third attempt at the same paragraph, the substantive question is going to the architect for arbitration: given that the source weighed comparable cost pressure and still concluded the stronger model should lead, does the recommendation survive, and what should the doc say? The fix round after that will have a settled answer to write rather than another guess.

---

sv-tmueller:
## Arbitration outcome — the Corollary conflict on #264

Dispatched to the architect rather than sending a third attempt at the same paragraph. **Ruling: the seat assignment survives; the argument carrying it does not and must narrow.** Settled — not to be re-raised in later rounds.

### The tester was right

`docs/reviews/2026-06-30-orchestration-comparison.md:126-133` opens "Cost cuts the other way" and prices Sonnet 5 at $3/$15 (intro $2/$10) against Opus 4.8's $5/$25 — a 1.67x to 2.5x premium for keeping the stronger model. This doc faces 2.0x. **That sits inside the band the Corollary faced and rejected.** The claim that the report's swap "had no offsetting pressure" is false, and the pressure was not even smaller. The two cases are epistemically identical too: the Corollary discounted its own cost case as unverified under the Max plan, which is exactly what section 8 concedes here.

### What the Corollary defeats, and what it does not

**Defeats:** the cost distinction entirely, and any framing in which this is capability-neutral or capability-positive.

**Does not reach:** availability — the Corollary contains no availability fact or argument anywhere. And the floor question: its conclusion in its own words is "keep Opus 4.8 as the default lead", which is an argument against dropping the lead *below Opus class*, not an argument that only the top-ranked model may lead.

### The honest distinction

Not "both are frontier tier, so the gap is smaller" — that leans on vendor ranking the doc itself calls its weakest evidence. The defensible argument is a floor claim checkable in this repo's own history: this repo ran an Opus-class lead by policy until `7fd326e` (#133) with no recorded inadequacy; #133 moved up because a stronger model appeared, not because Opus leadership failed; Opus 5 is that model's successor at identical pricing and is already this repo's documented judgment-seat fallback per `ec14b7a` on this very branch. **That argument needs only "Opus 5 >= Opus 4.8", not any claim about the Fable-to-Opus-5 gap.**

### A counterweight nobody had engaged

The Model policy opens "The strongest model in every plan/decision seat." Commit `7fd326e` (#133, one day after the Corollary) moved the lead to Fable with the reasoning that the lead "routes every handoff, refines batches, and makes the dispatch decisions, so it gets the strongest model." **The Corollary's principle was not superseded — it was ratified into standing policy.** A lead-only move to Opus 5 is an exception to that written line. It does not sink the recommendation, but the doc must admit it, and it makes the follow-up a human sign-off item rather than an in-scope decision.

### Corrections the ruling requires beyond the rebuttal

- Section 8's "removes that already-observed failure mode" is **false** and must go. Architect and reviewer stay on Fable and can still hit the limit. The real mechanism is reduced exhaustion pressure on a pool the lead plausibly dominates
- Both recorded events were **subagent judgment-seat dispatches**, not lead-seat blocks
- The summary's "one-command, zero-code" conflicts with section 7's own inventory: `README.md:77`, `docs/team-architecture.md:33`, and the team-guide orchestrator bullet all hardcode the lead as Fable. Cheap to reverse, not zero-edit

### Explicitly discounted

**Vendor "narrowing the gap":** the quoted sentence reaffirms "Claude Fable 5 remains the highest-capability tier" in the same breath. It makes no claim about the size of the Opus 5 to Fable 5 gap. Using it as gap-narrowing evidence misreads the source.

**Reversibility as a distinction from the Corollary:** the Corollary was arguing about a lead-model default too, equally one command. Reversibility lowers the bar only when harm is detectable, and degraded un-delegated lead judgment is precisely the class with no oracle. A modest plus, already priced into medium.

### Confidence: medium, unchanged

Both arguments to move it are rejected. The tester's finding should not push it below medium — the false claim was decorative, since section 8 already said the dollar range may not translate and that availability is what bites. The earlier "understated because reversible" should not push it to high — reversibility is already inside the medium call and discounted by poor detectability, and this arbitration surfaced two new counterweights (the policy exception, and #133 ratifying the Corollary's reasoning).

Off high: no first-party capability comparison, unmeasured quota effect, two events that are not a rate, and a policy exception needing human sign-off. Off low: an Opus-class floor this repo has real operating experience with, a recurring availability failure the policy itself names as a switch trigger, and cheap reversal.

Fix round dispatched with this ruling.

---

sv-tmueller:
## Tester verdict, round 4 — issue #264: PASS

Commit under test: `adb8ef5`. `npm test` 65/65. No must-fix findings.

### The reflow preserved content

`adb8ef5` shows 318 insertions / 318 deletions on a 417-line file because the whole document was re-wrapped to hold a line ceiling. The diff is therefore not reviewable as a delta, so this round verified the final state instead.

`git diff --word-diff` confirms the change is surgical despite the churn: every hunk is a targeted rewrite of section 8, plus two small phrasing fixes and the section 7 token-attribution fix. Nothing outside the arbitration's scope was touched, and no argument or citation was dropped — confirmed by word-diff and a full top-to-bottom read.

### The arbitration ruling is implemented

All five required elements present and verified against primary sources:

1. **Cost conceded outright** (lines 322-326), with 1.67x/2.5x checked against `2026-06-30-orchestration-comparison.md:126-136`
2. **Floor claim, not ranking claim** (327-336). `git show 248d672 -- CLAUDE.md` and `git show 7fd326e` both check out, including that #133 landed 2026-07-01, one day after the Corollary
3. **Availability with all three precisions** (339-344); `11003f9` verified
4. **The policy conflict named** (344-353), citing the team-guide's opening line and `7fd326e` as the ratifying commit, with the follow-up flagged a human sign-off item
5. **Capability concession kept** (353-356), vendor quote matching `shared/models.md` exactly

All five forbidden claims grepped and absent. Corrections A, B and C present and verified.

### Regressions — none

The self-verification distinction survived the rewrap verbatim, re-checked against `tm-review-changes.js:145` and both sibling workflows. The change-surface inventory, every cited figure, the hedged `opus` alias claim, the top summary, and the no-enactment framing all intact. Criterion 6 confirmed across all four commits.

### Nits

1. Line 328 quotes "keep Opus 4.8 as the default lead" as verbatim; the source reads "favors keeping Opus 4.8 as the default lead". Meaning survives, exactness does not
2. `adb8ef5`'s commit body references holding "the 200-300 line target", but the file is 417 lines both before and after. Stale reference to an earlier round's ceiling
3. Section 7 does not literally name the team-guide orchestrator bullet as a lead-seat hardcode; it covers team-guide generically as "the policy's own prose". The summary's reconciliation is directionally right

### Readability — the tester's opinion, which I asked for and think is correct

The top summary works: recommendation, confidence and blocker in ten seconds. But section 8, now carrying the entire arbitrated rebuttal, reads worse than the density metric suggests. Its three body paragraphs each run 150-220 words and pack three to five distinct logical moves into semicolon-chained sentences with inline citations. That is an argument built for an auditor tracing claims to commits, not for a human deciding from it in one read.

The cause is mine: I set a hard line ceiling, and it was met by parataxis rather than cutting. Line count was the wrong constraint for prose.

The tester's suggested fix, which I am forwarding to review rather than acting on unilaterally: break the rebuttal into the same five labeled items the arbitration used — cost, floor claim, availability, policy conflict, capability concession — as a short list rather than three run-on paragraphs. Costs a few lines, makes the most important section scannable.

### Untested claims

- The #262 record of the two Fable-limit events — no GitHub access in the agent sandbox, the same limitation the document itself discloses. **Lead confirmation: #262 records both events with the exact error string**
- The document's analytical claims have no automated coverage; `npm test` covers workflow code only. Verification was manual against primary sources
- The recommendation is a judgment call by design, which no test can confirm or refute

---

sv-tmueller:
## Reviewer verdict, round 2 — issue #264: APPROVE

Commit `adb8ef5`. Pass 1 clean: one file, no policy edit, no pin, no workflow stage, no A/B run, non-goals held. Pass 2: the argument holds and the sourcing is genuinely good.

Independently spot-verified as correct: the 93.1%/93.3% cross-check (and it is a real test — per-bucket weights differ, so bucket mix does move the answer), 81.9%, 40.9%, 43.6%, 3.33x, 10x, 2.0x, the $7.55/$6.44/$1.11/$1.80 package figures, the 288-of-297 addendum including the corrected token attribution, `effort-policy.test.mjs:21` and its exact error string, the three critic-stage line numbers, `README.md:77`, `docs/team-architecture.md:33`, the capability and pricing quotes, the alias table, and the behavioral-shifts range.

> The self-verification carve-out in section 7 is the sharpest thing in the document and is correct.

The arbitration is faithfully implemented: cost conceded, the floor claim carrying the argument, all three availability precisions, the #133 policy exception and its human sign-off named, "narrowing the gap" gone, confidence medium. No forbidden claim survives.

### Ruling on the section 8 restructure: should-fix, ship it

The tester's readability read is correct — three 150-220 word paragraphs plus a five-clause semicolon chain is an audit trail, not a decision aid, and CLAUDE.md's "short sentences" is on its side. It does not block: the actual decision aid is the accurate four-line summary; section 8's density produces no error (every claim in it traced correct); and the cause was a lead-imposed line ceiling, so fixing it is a re-spec rather than a defect correction. "If you run a round for anything else, fold this in first — it is the highest-value item on the list."

### Findings

**1. Lines 182, 342, 379 — the availability evidence is counted three different ways, and the sweep missed an event.** "Two events across two separate occasions" / "a recurrence across three occasions" / "two isolated incidents". Separately, `docs/architecture/2026-07-05-codebase-map.md:6` records a third occasion the document never cites: the 2026-07-04 map's synthesis stage "ran on the Opus 4.8 fallback while Fable 5 was quota-exhausted (#179)". The error is conservative — it understates a pillar of the recommendation.

**2. Lines 71, 232, 297 — a conditional test failure described as a current one.** "Today's failing effort-policy test", "the effort-policy failure below", "blocked today by a real, verified test failure". `npm test` at `adb8ef5` is green, 65 pass / 0 fail. Section 7's own statement is correctly conditional, and its evidence is a read of the assertion, not a run.

**3. Line 328 — the misquote**, as the tester found. This document stakes its value on quotes an auditor can diff.

**4. Lines 8-10 — the summary names the wrong leading argument.** It calls cheap reversibility "the strongest argument for acting despite the unmeasured capability risk", but reversibility is argued nowhere in section 8, and the arbitration priced it as a modest plus already inside medium. Section 8's actual arguments are the Opus-class floor, the bounded savings, and availability.

**5. Lines 290-369 — the section 8 restructure** (the ruling above).

### Nits

6. Line 236 cites test body lines 37-61 for both the frontmatter and workflow-stage failures; 37-61 is only the frontmatter block, the stage assertion is at 86-105. True claim, half-covered citation.
7. Line 331 reads as if `248d672` were the replacing commit; it is #87, three weeks earlier, cited as evidence of the pre-#133 lead bullet.
8. Lines 109-110 replace the source's parentheses with a comma in the Sonnet pricing quote. Punctuation only.

### Lead decision

Taking the round. Findings 1 and 2 are accuracy defects rather than polish — a document that contradicts itself three ways about its own key evidence, and states a passing test is failing, undermines the trust its citation discipline is meant to earn. Both are cheap. Per the reviewer's own guidance, findings 3-8 fold in, including the section 8 restructure, and the line ceiling is lifted since it caused the density.

### Untested claims, forwarded

- The #262 record of the Fable-limit events — no GitHub access in the agent sandboxes. **Lead confirmation: #262 records both with the exact error string.** Note finding 1 means the true count is higher than the document currently claims
- The document's analytical claims have no automated coverage; verification was manual against primary sources
- The recommendation is a judgment call no test can confirm

---

sv-tmueller:
## Final round verified — PR marked ready for review

Commit `6f73e2d` on top of the reviewer's APPROVE. `npm test` 65/65 on the final commit. One file touched, 125 insertions / 100 deletions. Final length 442 lines (the ceiling was lifted for this round, since it caused the density problem).

Lead verification of the two accuracy defects that justified taking the round:

| Check | Result |
|---|---|
| Availability count reconciled | **"Four events across three separate documented occasions"**, stated identically everywhere it appears. Two dispatches this batch (#262), the prior occurrence behind `11003f9` (#228), and the 2026-07-04 codebase-map synthesis fallback (#179) |
| #179 now cited | Yes, 4 references |
| Present-tense failing-test claims | Both strong assertions gone; conditional "would fail" phrasing in place |
| Misquote | Fixed — now matches `2026-06-30-orchestration-comparison.md:135` verbatim |
| Only the research doc touched | Confirmed |

Findings 4-8 also reported fixed, including the section 8 restructure into the arbitration's five labels with every clause, number and citation traced across.

### One residual nit, shipping as-is

Line 241 still reads "the effort-policy failure below" where "blocker" would be accurate. The substantive defect is fixed — the two assertions that the suite is currently red are gone, and lines 74 and 308 are properly conditional — so no reader will conclude `npm test` is failing. It is a one-word residue.

I have not edited it myself. The value of this pipeline all batch has been that nobody verifies their own work, and a lead patching the artifact after approval to avoid a round would undercut exactly that. Correct it at merge if it bothers you.

### Package history

- **#263:** tester PASS round 1, reviewer APPROVE round 1. Zero findings against the diff
- **#264:** tester FAIL -> PASS -> FAIL -> arbitration -> PASS; reviewer CHANGES_REQUESTED -> APPROVE. Four commits, one architect arbitration, one fact-checker audit, one recovered commit

Ready for human review.

---

Issue #266, opened 2026-07-25
Batch: fable-outage-hardening


sv-tmueller:
Approved contract for this batch, verbatim from the /tm-advisor sign-off. Dispatched 2026-07-25. Follows batch #262 (closed, PR #265 merged).

## Backlog scan that produced this batch

Three open issues. Two are not dispatchable:

- **#232** (Codex spike) — its own body says "Runs on the owner's machine with real subscription logins. This is not a headless-agent task; do not dispatch it through the kickoff pipeline." Excluded by instruction.
- **#224** (re-measure token burn) — its date gate (2026-07-20) has passed, but it is not runnable in a container. The environment has exactly one transcript directory (`-home-user-orchestrai`) against a baseline built from a 14-day, 25-project aggregate. Running it would compare a new window to 1/25th of the data and label the result a delta. Needs the owner's machine.
- **#261** (auto-recover a Fable-quota death onto Opus) — dispatchable, sliced below.

Correction to #261's framing, for whoever works it: it quotes the team guide as "Fallback: Opus 4.8 at xhigh", which is stale as of PR #265. One of its stated obstacles has also moved — `089f754` made the pin-flip pass tests. Its remaining obstacle (plugin-cache edits get overwritten on update) still stands.

---

### 1. Auto-recover the Fable critic stage onto Opus 5 in the tm- workflows

**Scope.** When the Fable-critic stage of `tm-review-changes`, `tm-review-codebase` or `tm-map-codebase` dies on a Fable quota exhaustion, the workflow retries that stage once on Opus 5 at the same effort and records the switch in its output, instead of losing the whole run. This is the half of #261 that matters most: workflow scripts run unattended with no lead watching, so a critic death there fails the entire run, whereas a lead-level dispatch already has a documented manual recovery that works.

One implementation note that will otherwise cost a round: the Workflow `agent()` call **returns `null`** on a terminal API error rather than throwing, so the retry is a null check, not a `try`/`catch`. It also returns `null` when a user skips the agent mid-run, so the package has to decide whether an unconditional retry is acceptable or the two cases need distinguishing.

**Acceptance criteria**
- [ ] A Fable-critic stage that returns `null` is retried once on Opus 5 at the same pinned effort, in all three workflows
- [ ] The retry and its reason appear in the workflow's user-visible output, so a silent model switch is impossible
- [ ] The ambiguity between a quota death and a user skip is explicitly decided and documented in the code comment, not left implicit
[...]

Issue #261, opened 2026-07-24
kickoff/advisor: auto-recover a judgment-role dispatch that dies on Fable 5 quota onto Opus


sv-tmueller:
## Summary

During a Fable 5 quota outage, every judgment-role dispatch (architect, reviewer, and the Fable-critic stage of the tm- workflows) dies once with a terminal API error before the orchestrator can recover it. The documented Opus fallback works, but only after a hard agent death and only if the orchestrator is watching. Automate that recovery, or provide a batch-level switch, so a Fable outage stops producing a hard failure per dispatch.

## What happens today

The judgment seats pin `model: fable`:

- `agents/reviewer.md` and `agents/architect.md` both carry `model: fable` with the comment `# Judgment role: Fable 5. Fallback is opus (see team-guide, Model policy).`

The team-guide Model policy states the fallback is deliberately manual:

> Fallback: Opus 4.8 at xhigh effort (a procedure, not automatic). When Fable 5 is unavailable, rate-limited, quota-exhausted, or refuses the workload ... For a single role-agent dispatch hitting Fable quota mid-batch, dispatch that one agent with a per-call Opus override (the Agent tool's `model` param) instead of flipping every pin.

Observed on 2026-07-24, mid-batch under `/tm-advisor`: a reviewer dispatch terminated early with

> Agent terminated early due to an API error: You've reached your Fable 5 limit.

zero output, no verdict. The orchestrator recovered exactly as the team-guide prescribes: re-dispatched that one agent with a per-call Opus override, which returned APPROVE. So the fallback is not broken. The cost is that the agent has to die first, and a human/orchestrator has to notice and retry.

## Why it is worth smoothing

The manual procedure is fine for one stray dispatch. It scales badly during a sustained Fable outage (the exact scenario in `Cost-based fallback trigger` and `unavailable / rate-limited / quota-exhausted`):

- Fable is quota-exhausted for the rest of a billing window, so it is not one dispatch, it is *every* judgment dispatch in the batch (each reviewer, each architect arbitration, each Fable-critic workflow stage).
- Each one costs a full agent spin-up that ends in a terminal error before recovery.
- An unattended or dropped-and-resumed run can stall on it if nobody is watching to perform the manual override.
- The clean documented alternative (flip the `fable` pins to `opus` in the agent frontmatters and workflow critic stages) edits files in the plugin cache, which the next plugin update overwrites.

## Proposal (any one)

[...]

PR #270, opened 2026-07-25
Harden the Fable outage path: workflow auto-recovery, first CI, and a research-doc addendum

### LINKED ISSUE #267: Auto-recover the Fable critic stage onto Opus 5 in the tm- workflows
Part of batch #266

Implements the workflow half of #261.

## What to build

When the Fable-critic stage of `tm-review-changes`, `tm-review-codebase` or `tm-map-codebase` dies because the Fable 5 quota is exhausted, the workflow retries that stage once on Opus 5 at the same effort and records the switch in its output, instead of losing the whole run.

This is the half of #261 that matters most. Workflow scripts run unattended with no lead watching, so a critic death there fails the entire run. A lead-level dispatch already has a documented manual recovery that works — batch #262 used it twice successfully.

One implementation note that will otherwise cost a round: the Workflow `agent()` call **returns `null`** on a terminal API error rather than throwing, so the retry is a null check, not a `try`/`catch`. It also returns `null` when a user skips the agent mid-run, so this package has to decide whether an unconditional retry is acceptable or the two cases need distinguishing, and say which in the code.

Context that has moved since #261 was filed: the team-guide text it quotes ("Fallback: Opus 4.8 at xhigh") is stale as of PR #265, and `089f754` added an `opus` tier to the effort-policy test so an Opus stage now passes the suite at `xhigh`.

## Acceptance criteria

- [ ] A Fable-critic stage that returns `null` is retried once on Opus 5 at the same pinned effort, in all three workflows
- [ ] The retry and its reason appear in the workflow's user-visible output, so a silent model switch is impossible
- [ ] The ambiguity between a quota death and a user skip is explicitly decided and documented in the code comment, not left implicit
- [ ] If the Opus retry also returns `null`, the workflow fails with a clear message rather than proceeding with a missing critic
- [ ] The retry helper is byte-identical across all three workflows, satisfying the existing `helpers.test.mjs` invariant
- [ ] A test covers the retry path
- [ ] `npm test` passes

## Non-goals

- No change to which model a stage runs by default; Fable 5 stays primary
- No change to any `effort:` pin
- No auto-recovery for lead-level agent dispatches (that path already has a working documented procedure)
- Does not address the plugin-cache-overwrite problem #261 raises

### LINKED ISSUE #268: Give this repo the CI its own team guide specifies
Part of batch #266

## What to build

`npm test` runs as an independent gate on pull requests and on pushes to `main`.

The team guide's CI cost policy specifies pinned `timeout-minutes`, a mandatory `concurrency` group with `cancel-in-progress`, staged jobs, and a draft-PR skip for expensive stages. It ships a worked template at `.claude/skills/tm-new-project/templates/ci.yml`. But this repo has no `.github/` directory at all — it is the one repo its own policy does not apply to. Every check that has run here has been an agent or a human remembering to run `npm test` locally, with nothing independent confirming it.

Adapt the template to what this repo actually has. There is no application runtime: no install step, no typecheck, no lint, no build, no e2e. `package.json` defines exactly one script, `node --test ".claude/workflows/__tests__/*.test.mjs"`, which runs 65 tests in about 240ms. The result should be a single small job, not the template's three-stage pipeline with placeholder commands.

## Acceptance criteria

- [ ] `.github/workflows/ci.yml` runs `npm test` on pull requests and on pushes to `main`
- [ ] The job pins `timeout-minutes` and the workflow declares a `concurrency` group with `cancel-in-progress: true`
- [ ] Template stages this repo has no equivalent for (install, typecheck, lint, build, e2e) are removed rather than left as placeholder `echo` commands
- [ ] The template's `paths-ignore` trade-off is either kept with its known blocked-required-check caveat noted, or dropped with a one-line reason
- [ ] The workflow is valid YAML and its `run:` command matches this repo's actual test script

## Non-goals

- No new dependency, no test-framework change, no new npm script
- Does not add e2e or build stages this repo has nothing to put in
- Does not make the repo public or change its visibility
- Does not modify the `tm-new-project` template itself

### LINKED ISSUE #269: Add a dated addendum to the Opus 5 research doc
Part of batch #266

## What to build

`docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md` recommends holding architect, reviewer and the three workflow critic stages on Fable 5 **because the effort-policy test blocked flipping their pins to `opus`**. That blocker was cleared by commit `089f754` a few hours after the doc was written. A future reader will act on a constraint that no longer exists.

Append a dated addendum recording that the blocker is gone, and what that does and does not change. It does not overturn the recommendation: the second reason for holding those seats — vendor positioning still ranks Fable 5 above Opus 5 on capability, with no first-party comparison available — is untouched and stands on its own.

The doc body stays as written. It is a dated record of the analysis as performed on 2026-07-24, and rewriting dated findings after the fact is the error batch #262 explicitly declined to make with the repo's historical Opus 4.8 references. An addendum is the right shape for correcting a dated record.

One residual wording nit found in review and disclosed rather than silently patched: line 241 says "the effort-policy failure" where "blocker" is accurate. That line is now doubly stale, since the blocker is fixed. Fix the wording as part of this package.

## Acceptance criteria

- [ ] A dated addendum section records that `089f754` cleared the effort-policy blocker and links it
- [ ] The addendum states plainly which part of the recommendation is affected (the hold on architect, reviewer and the critic stages) and which is not (the lead-seat move, and the capability argument)
- [ ] The original analysis text is unchanged apart from the line 241 wording fix
- [ ] Line 241 reads "blocker" rather than "failure"
- [ ] `npm test` passes

## Non-goals

- Does not revise the recommendation, its confidence level, or any figure
- Does not rewrite the body of the analysis
- Does not act on the recommendation or touch `.claude/team-guide.md`

sv-tmueller:
Batch #266, run through `/tm-advisor`. Three packages, one branch - this session is pinned to a single branch, so all three land here as separate commits instead of one PR each. Rationale logged on #266.

Closes #267
Closes #268
Closes #269

## Why

Batch #262 hit the Fable 5 quota limit twice and recovered manually both times. Dispatching this batch hit it three more times: **every** judgment dispatch failed on quota, two days running. That is seven events across four documented occasions, and it is no longer the stray-dispatch nuisance #261 was filed about. It is the sustained-outage case its manual procedure was explicitly not written for.

## What changed

Nine commits across three packages.

**#268 - the first CI this repo has ever had** (`c3efac4`, `a0b8289`)

The team guide's CI cost policy specifies pinned timeouts, mandatory concurrency groups and staged jobs, and ships a worked template. This repo had no `.github/` directory at all. Every check that ever ran here was an agent or a human remembering to run `npm test` locally.

Adds one job: checkout, `actions/setup-node@v4` pinned to Node 22, `npm test`. Five-minute timeout, concurrency group with `cancel-in-progress`, `permissions: contents: read`.

Three decisions worth knowing:

- **`paths-ignore` dropped.** 47 of 61 tracked files here are markdown. Three of the last five merges touched only `docs/` or `.claude/*.md`, so the template's version would have skipped CI on all three, and a docs-only PR would hang forever as "Expected" once the check is required
- **Node pinned, and it is load-bearing.** The test script is `node --test ".claude/workflows/__tests__/*.test.mjs"` with quotes inside the script string, so **Node itself expands the glob**. `ubuntu-latest` ships whatever Node its image pins, and that moves without a repo change
- **No `cache: 'npm'` and no install step.** There is no lockfile. `npm ci` would fail hard and the cache step would too. Commented so nobody adds it later

**#267 - auto-recover the Fable critic onto Opus 5** (`6049fb1`, `41ab06f`, `9e011e3`)

`criticWithFallback`, triplicated byte-identically across the three `tm-` workflows. A critic stage that returns `null` retries once on Opus 5 at the same `xhigh` pin; if the retry also returns `null` the workflow throws rather than proceeding with a missing critic.

The retry is a null check, not a `try`/`catch`, because `agent()` returns `null` on a terminal API error rather than throwing. It also returns `null` when a user skips the dispatch, and nothing in the script can tell those apart. The code rules on asymmetric cost and says so: retrying a deliberate skip costs one extra skip prompt, not retrying a quota death costs the entire unattended run. Retry unconditionally.

**#269 - dated addendum to the Opus 5 research doc** (`328a8fe`, `5cc6646`, `47373af`)

`docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md` held architect, reviewer and the critic stages on Fable partly because the effort-policy test blocked flipping their pins to `opus`. `089f754` cleared that. The addendum records it without rewriting the dated analysis, and states plainly what it does not overturn: the capability argument for holding those seats is untouched and stands on its own.

## Two corrections to the record

**A commit message in this branch overclaims.** `5cc6646`'s body says it "recomputed every citation against the file as delivered and verified each with `sed -n`". That was not true of that commit: `47373af` exists because two citations were missed, one of which had been broken since `328a8fe` and never disclosed. `47373af`'s own message corrects it, naming both misses with their wrong and right targets. Recorded here rather than rewritten, since rewriting would need a force-push.

**A false claim reached three files before being caught.** `41ab06f` added a comment stating that workflow scripts are ES modules and therefore strict-mode code. That came from a lead instruction, verbatim. It is wrong: these files carry `export const meta` on line 1 and a bare top-level `return`, a combination that is a SyntaxError as a module and as a script, so the file cannot tell you its own mode. `9e011e3` removes it and rests the rationale on the mechanism that actually applies: mutating a frozen return silently drops the marker in sloppy mode. Confirmed by reverting to the mutation form and observing `modelFallback` come back `undefined` rather than throwing, and by zero `use strict` matches anywhere in `.claude/workflows/`.

## Known inherited risk, not introduced here

`node --test` with a glob matching nothing exits **0** with `# tests 0`. A green CI run is therefore not by itself proof that tests ran. This is a pre-existing property of the repo's `npm test` script, and fixing it needs an npm script change #268's non-goals forbid. Today the glob matches all 70 tests. Worth knowing before anyone moves the test files.

## What the CI found by running on itself

**A green run confirmed `# tests 70`, not 0.** The known hazard above did not fire.

**A run can fail with zero jobs.** `41ab06f` concluded `failure` having created no jobs at all, while `ci.yml` was byte-identical to the green run on its parent and `npm test` passed locally at that exact commit. Re-running produced no jobs either; the next commit ran green. Transient, but it means a red X here does not always mean a test failed.

**The action deprecation escalated mid-batch.** The first run warned that `actions/checkout@v4` and `actions/setup-node@v4` target the deprecated Node 20. A later run reports them "being forced to run on Node.js 24". That is a live compatibility risk, not housekeeping.

## After merge, two things only you can do

**The workflow existing is not the workflow gating.** Making "Unit tests" a required check is a branch-protection setting, not a file, and no package in this batch can do it. Until you flip it, this CI is advisory.

**Bump the actions to v5 before you flip it.** A required check that can conclude `failure` with zero jobs would block merges on infrastructure noise, and the checks UI does not distinguish that from a real failure.

## Verification

- `npm test` 70 pass, 0 fail; CI green on `9e011e3` with the count confirmed from the run log
- **#268:** tester PASS round 1, reviewer APPROVE, zero findings, including an independent recount of the 77% markdown figure
- **#267:** tester PASS -> reviewer CHANGES_REQUESTED -> PASS -> APPROVE -> PASS -> APPROVE. The blocking finding was a wiring test that could not fail: its assertion was satisfied by the function declaration itself, so nothing verified the helper was ever called
- **#269:** tester FAIL -> FAIL -> PASS, reviewer APPROVE. Both failures were the same defect, line-number citations pointing at unrelated prose, and both were found only when the tester enumerated the document itself instead of checking a handed-over list
- Every verification claim in this PR was re-derived by an agent that did not make it: byte-identity by independent extraction rather than by the test that asserts it, both deliberate breaks reproduced on different files than the developer used, and the retry's failure mode observed rather than inferred

Non-blocking findings from all three packages are in the comments for your review; none were silently fixed.

https://claude.ai/code/session_01QMT3YhDmHtzrnTkfTwVWth

---
_Generated by [Claude Code](https://claude.ai/code/session_01QMT3YhDmHtzrnTkfTwVWth)_

### DIFF
--- a/.claude/team-guide.md
+++ b/.claude/team-guide.md
@@ -196,8 +196,9 @@ Rationale: docs/team-guide-rationale.md.
   Fable 5 is unavailable, rate-limited, quota-exhausted, or
   refuses the workload, switch the lead with `/model claude-opus-5`; for a
   longer outage flip the `fable` pins to `opus` in the `architect` and
-  `reviewer` frontmatters and the Fable-critic stage of every `tm-` workflow
-  (`.claude/workflows/`). Flip back when Fable returns. For a single
+  `reviewer` frontmatters (the Fable-critic stage of every `tm-` workflow
+  already retries on opus automatically when Fable returns nothing, no flip
+  needed there; `.claude/workflows/`). Flip back when Fable returns. For a single
   role-agent dispatch hitting Fable quota mid-batch, dispatch that one agent
   with a per-call Opus override (the Agent tool's `model` param) instead of
   flipping every pin. The ladder is Fable -> Opus; Sonnet is never a fallback
--- a/.claude/team-guide.md
+++ b/.claude/team-guide.md
@@ -196,12 +196,14 @@ Rationale: docs/team-guide-rationale.md.
   Fable 5 is unavailable, rate-limited, quota-exhausted, or
   refuses the workload, switch the lead with `/model claude-opus-5`; for a
   longer outage flip the `fable` pins to `opus` in the `architect` and
-  `reviewer` frontmatters (the Fable-critic stage of every `tm-` workflow
-  already retries on opus automatically when Fable returns nothing, no flip
-  needed there; `.claude/workflows/`). Flip back when Fable returns. For a single
-  role-agent dispatch hitting Fable quota mid-batch, dispatch that one agent
-  with a per-call Opus override (the Agent tool's `model` param) instead of
-  flipping every pin. The ladder is Fable -> Opus; Sonnet is never a fallback
+  `reviewer` frontmatters. The Fable-critic stage of every `tm-` workflow
+  (`.claude/workflows/`) already retries on opus automatically when Fable
+  returns nothing, so no flip is needed there, though each run still burns
+  one doomed Fable dispatch before the retry fires. Flip back when Fable
+  returns. For a single role-agent dispatch hitting Fable quota mid-batch,
+  dispatch that one agent with a per-call Opus override (the Agent tool's
+  `model` param) instead of flipping every pin. The ladder is Fable -> Opus;
+  Sonnet is never a fallback
   for a judgment seat, and only steps in, flagged, if Opus is also down, with
   the review re-run on a judgment model afterward.
 - Cost-based fallback trigger: if Fable 5 stops being included under the
--- a/.claude/workflows/__tests__/effort-policy.test.mjs
+++ b/.claude/workflows/__tests__/effort-policy.test.mjs
@@ -71,7 +71,11 @@ describe('agent frontmatter effort pins', () => {
 //
 // Agent-call opts in both workflows are single-line objects carrying both
 // label: and model:. The meta.phases display entries carry model: but
-// title:/detail: instead of label:, so requiring label: excludes them.
+// title:/detail: instead of label:, so requiring label: excludes them. The
+// criticWithFallback retry opts (`{ ...opts, model: 'opus' }`) also carry
+// model: with no label: and no literal effort:, so the same filter excludes
+// them too; that is safe because the spread inherits effort: from the
+// caller's opts line, which this test already checks.
 // ===========================================================================
 describe('workflow stage effort pins', () => {
   for (const file of WORKFLOW_FILES) {
--- a/.claude/workflows/__tests__/helpers.test.mjs
+++ b/.claude/workflows/__tests__/helpers.test.mjs
@@ -482,9 +482,14 @@ describe('criticWithFallback', () => {
   })
 
   test('every workflow file wires the critic call through criticWithFallback', () => {
+    // Assert on 'await criticWithFallback(', not the bare function name: the
+    // declaration line ('async function criticWithFallback(...') also matches
+    // the bare name, so that substring is present even if the call site were
+    // reverted to `await agent(`. The declaration has no `await`, so this
+    // string only matches an actual call.
     for (const file of FILES) {
       const src = readFileSync(join(workflowsDir, file), 'utf8')
-      assert.ok(src.includes('criticWithFallback('), `${file} does not call criticWithFallback(`)
+      assert.ok(src.includes('await criticWithFallback('), `${file} does not call criticWithFallback(`)
     }
   })
 })
--- a/.claude/workflows/__tests__/helpers.test.mjs
+++ b/.claude/workflows/__tests__/helpers.test.mjs
@@ -10,6 +10,10 @@
  * union in tm-review-codebase.js. Those are tested as logic kernels - the
  * test captures the expression's shape, not the real source text. This is
  * the one honest seam; it is documented below and called out in the PR.
+ *
+ * loadFn() accepts an optional sandbox object so a function whose free
+ * variables are workflow runtime globals (agent, log) can be run with stub
+ * implementations instead of throwing ReferenceError.
  */
 
 import { test, describe } from 'node:test'
@@ -27,10 +31,12 @@ const workflowsDir = join(__dir, '..')
 // Helper: slice a named function out of a JS source file and evaluate it in a
 // vm context. Returns the function value.
 // ---------------------------------------------------------------------------
-function loadFn(filename, fnName) {
+function loadFn(filename, fnName, sandbox = {}) {
   const src = readFileSync(join(workflowsDir, filename), 'utf8')
-  // Match "function <fnName>(" - handles single-line bodies too.
-  const startRe = new RegExp(`function ${fnName}\\s*\\(`)
+  // Match "function <fnName>(", optionally preceded by "async" - handles
+  // single-line bodies too. The async keyword must stay inside the slice, or
+  // runInContext throws on the function's first await.
+  const startRe = new RegExp(`(?:async\\s+)?function ${fnName}\\s*\\(`)
   const startMatch = startRe.exec(src)
   if (!startMatch) throw new Error(`${fnName} not found in ${filename}`)
 
@@ -46,7 +52,7 @@ function loadFn(filename, fnName) {
   }
   const fnSrc = src.slice(startMatch.index, pos)
 
-  const ctx = createContext({})
+  const ctx = createContext(sandbox)
   runInContext(fnSrc, ctx)
   return runInContext(fnName, ctx)
 }
@@ -56,7 +62,7 @@ function loadFn(filename, fnName) {
 // ---------------------------------------------------------------------------
 function sliceFnSrc(filename, fnName) {
   const src = readFileSync(join(workflowsDir, filename), 'utf8')
-  const startRe = new RegExp(`function ${fnName}\\s*\\(`)
+  const startRe = new RegExp(`(?:async\\s+)?function ${fnName}\\s*\\(`)
   const startMatch = startRe.exec(src)
   if (!startMatch) throw new Error(`${fnName} not found in ${filename}`)
   let pos = startMatch.index
@@ -384,5 +390,101 @@ describe('scoutDropped union', () => {
       mapDropped: null,
     })
     assert.deepEqual(result, [])
+  })
+})
+
+// ===========================================================================
+// 5. criticWithFallback (issue #267)
+//
+// The helper's only free variables are the workflow runtime globals `agent`
+// and `log`. Supplying stubs as the vm sandbox runs the actual shipped code,
+// not a hand-copied version.
+// ===========================================================================
+describe('criticWithFallback', () => {
+  const FILES = ['tm-review-changes.js', 'tm-review-codebase.js', 'tm-map-codebase.js']
+  const baseOpts = { label: 'consolidate', phase: 'Consolidate', model: 'fable', effort: 'xhigh', schema: { type: 'object' } }
+
+  test('happy path: first agent call succeeds, result returned unchanged', async () => {
+    const calls = []
+    const logs = []
+    const sandbox = {
+      agent: async (prompt, opts) => {
+        calls.push({ prompt, opts })
+        return { verdict: 'approve' }
+      },
+      log: (msg) => logs.push(msg),
+    }
+    const criticWithFallback = loadFn('tm-review-changes.js', 'criticWithFallback', sandbox)
+
+    const result = await criticWithFallback('the prompt', baseOpts)
+
+    assert.deepEqual(result, { verdict: 'approve' })
+    assert.equal(calls.length, 1)
+    assert.equal(logs.length, 0)
+    assert.equal('modelFallback' in result, false)
+  })
+
+  test('first call returns null, retries on opus and succeeds', async () => {
+    const calls = []
+    const logs = []
+    const sandbox = {
+      agent: async (prompt, opts) => {
+        calls.push({ prompt, opts })
+        return calls.length === 1 ? null : { verdict: 'approve' }
+      },
+      log: (msg) => logs.push(msg),
+    }
+    const criticWithFallback = loadFn('tm-review-changes.js', 'criticWithFallback', sandbox)
+
+    const result = await criticWithFallback('the prompt', baseOpts)
+
+    assert.equal(calls.length, 2)
+    const secondOpts = calls[1].opts
+    assert.equal(secondOpts.model, 'opus')
+    assert.equal(secondOpts.effort, baseOpts.effort)
+    assert.equal(secondOpts.label, baseOpts.label)
+    assert.equal(secondOpts.phase, baseOpts.phase)
+    assert.equal(secondOpts.schema, baseOpts.schema)
+    assert.ok(calls[1].prompt.includes('the prompt'), 'second prompt still carries the original prompt')
+    assert.ok(calls[1].prompt.length > 'the prompt'.length, 'second prompt carries an added notice')
+    assert.equal(result.verdict, 'approve')
+    assert.equal(result.modelFallback, 'fable -> opus')
+    assert.ok(logs.length >= 1, 'log() must be called to surface the fallback')
+  })
+
+  test('both calls return null: throws naming the label and both models', async () => {
+    const sandbox = {
+      agent: async () => null,
+      log: () => {},
+    }
+    const criticWithFallback = loadFn('tm-review-changes.js', 'criticWithFallback', sandbox)
+
+    let threw = null
+    try {
+      await criticWithFallback('the prompt', baseOpts)
+    } catch (err) {
+      threw = err
+    }
+    assert.ok(threw, 'expected criticWithFallback to throw when both attempts return null')
+    // vm-realm errors are not `instanceof` the host's Error, so assert on the
+    // message rather than the error's prototype chain.
+    assert.match(threw.message, /consolidate/)
+    assert.match(threw.message, /fable/)
+    assert.match(threw.message, /opus/)
+  })
+
+  test('is byte-identical across all three workflow files', () => {
+    const a = sliceFnSrc('tm-review-changes.js', 'criticWithFallback')
+    const b = sliceFnSrc('tm-review-codebase.js', 'criticWithFallback')
+    const c = sliceFnSrc('tm-map-codebase.js', 'criticWithFallback')
+    assert.equal(a, b, 'criticWithFallback diverged between tm-review-changes.js and tm-review-codebase.js')
+    assert.equal(b, c, 'criticWithFallback diverged between tm-review-codebase.js and tm-map-codebase.js')
+  })
+
+  test('every workflow file wires the critic call through criticWithFallback', () => {
+    for (const file of FILES) {
+      const src = readFileSync(join(workflowsDir, file), 'utf8')
+      assert.ok(src.includes('criticWithFallback('), `${file} does not call criticWithFallback(`)
+    }
   })
 })
--- a/.claude/workflows/__tests__/helpers.test.mjs
+++ b/.claude/workflows/__tests__/helpers.test.mjs
@@ -430,7 +430,7 @@ describe('criticWithFallback', () => {
     const sandbox = {
       agent: async (prompt, opts) => {
         calls.push({ prompt, opts })
-        return calls.length === 1 ? null : { verdict: 'approve' }
+        return calls.length === 1 ? null : Object.freeze({ verdict: 'approve' })
       },
       log: (msg) => logs.push(msg),
     }
--- a/.claude/workflows/tm-map-codebase.js
+++ b/.claude/workflows/tm-map-codebase.js
@@ -81,10 +81,10 @@ async function criticWithFallback(prompt, opts) {
   )
   if (!second) throw new Error(`${opts.label}: both the ${opts.model} critic and the opus fallback returned nothing.`)
   log(`${opts.label}: this critique was produced by the opus fallback, not ${opts.model}.`)
-  // Return a new object rather than mutating `second`: workflow scripts are ES
-  // modules (strict mode), and the runtime's returned object could be frozen,
-  // sealed, or proxied, which would throw a TypeError on exactly this
-  // fallback path.
+  // Return a new object rather than mutating `second`: the runtime's returned
+  // object could be frozen or sealed, in which case mutating it would either
+  // silently drop the `modelFallback` marker (sloppy mode) or throw a
+  // TypeError, and a proxied result could throw either way.
   return { ...second, modelFallback: `${opts.model} -> opus` }
 }
 
--- a/.claude/workflows/tm-map-codebase.js
+++ b/.claude/workflows/tm-map-codebase.js
@@ -81,8 +81,11 @@ async function criticWithFallback(prompt, opts) {
   )
   if (!second) throw new Error(`${opts.label}: both the ${opts.model} critic and the opus fallback returned nothing.`)
   log(`${opts.label}: this critique was produced by the opus fallback, not ${opts.model}.`)
-  second.modelFallback = `${opts.model} -> opus`
-  return second
+  // Return a new object rather than mutating `second`: workflow scripts are ES
+  // modules (strict mode), and the runtime's returned object could be frozen,
+  // sealed, or proxied, which would throw a TypeError on exactly this
+  // fallback path.
+  return { ...second, modelFallback: `${opts.model} -> opus` }
 }
 
 const AREA = {
--- a/.claude/workflows/tm-map-codebase.js
+++ b/.claude/workflows/tm-map-codebase.js
@@ -15,8 +15,10 @@ export const meta = {
 // but never exceeds MAX_AREAS + 2, no matter how large the repo is or how many
 // areas the scout proposes. There is no per-file fan-out and no loop. Models and
 // effort are pinned per stage, so the session model and effort never leak into
-// the scout or the workers; the single critic runs Fable 5 at xhigh effort (flip
-// its model pin to 'opus' under the Opus 5 fallback, see team-guide).
+// the scout or the workers; the single critic runs Fable 5 at xhigh effort and
+// auto-retries once on opus at the same effort if Fable returns nothing
+// (criticWithFallback below; see team-guide for the lead-level fallback,
+// which is still manual).
 //
 // When the repo is too big for MAX_AREAS areas to cover, the leftover paths are
 // reported (coverage.areasDropped, coverage.ceilingReached, and a suggested next
@@ -55,6 +57,33 @@ function safeRef(value, fallback) {
 }
 const root = safeRef(opts.path, '.')
 const MAX_AREAS = Number.isInteger(opts.areas) && opts.areas > 0 ? opts.areas : 24
+
+// Duplicated byte-for-byte across the three tm- workflows that run a Fable
+// critic (the workflow runtime has no shared imports); keep this copy in
+// sync, see helpers.test.mjs.
+async function criticWithFallback(prompt, opts) {
+  const first = await agent(prompt, opts)
+  // agent() returns null both when the model errors out after retries (for
+  // example Fable 5 quota exhaustion) and when the user skips the dispatch
+  // mid-run. Nothing in this script can tell those two cases apart, so there
+  // is no heuristic to invent here. The ruling is on which failure mode is
+  // worse: retrying a deliberate skip costs one extra skip prompt (the retry
+  // is itself an agent() call, so skipping again just returns null and the
+  // run ends below with a clear error), while not retrying a quota death
+  // costs the entire unattended run. Retry unconditionally.
+  if (first) return first
+  log(
+    `${opts.label}: the ${opts.model} critic returned nothing. This could be quota exhaustion or a manual skip; retrying once on opus at the same effort. Skipping again will end the run.`
+  )
+  const second = await agent(
+    `${prompt}\n\nNOTE: this is a retry on the opus fallback after the ${opts.model} critic returned nothing (quota or a skip). If you write a report file, record in it that this fallback produced it.`,
+    { ...opts, model: 'opus' }
+  )
+  if (!second) throw new Error(`${opts.label}: both the ${opts.model} critic and the opus fallback returned nothing.`)
+  log(`${opts.label}: this critique was produced by the opus fallback, not ${opts.model}.`)
+  second.modelFallback = `${opts.model} -> opus`
+  return second
+}
 
 const AREA = {
   type: 'object',
@@ -231,7 +260,7 @@ const suggestedNextActionClause = ceilingReached
   ? `, and coverage.suggestedNextAction to ${JSON.stringify(suggestedNextAction)}`
   : ''
 
-const report = await agent(
+const report = await criticWithFallback(
   `You are the senior architect synthesizing a full-codebase map from the area descriptions below. This is a map, not a review: do not add findings, severities, recommendations, or quality and security judgments. Describe what is there.${coverageNote}\n\nRun \`date +%F\` for today's date, make the docs/architecture/ directory if it does not exist, and write docs/architecture/<date>-codebase-map.md with exactly these sections, in this order:\n1. Executive summary: what the repo is and how its areas fit together, in a few paragraphs.\n2. Component table: one row per area, with columns for area, purpose, and key modules.\n3. Data-flow narrative: how data moves across areas, end to end.\n4. Dependency summary: external dependencies (packages, services) and cross-area dependencies.\n5. Open questions: anything the workers could not determine or that needs a human to confirm.\n\nIf coverage is partial, add a "Coverage" callout immediately after the executive summary stating how many paths were not mapped and the suggested next action.\n\nSet reportPath to the file you wrote. Set summary to a short plain-language summary of the whole repo. Set openQuestions to the same list you put in the report's Open Questions section.\n\nSet coverage.areasMapped to ${JSON.stringify(mappedAreas)}, coverage.areasDropped to ${JSON.stringify(scoutDropped)}, coverage.workersFailed to ${JSON.stringify(workersFailed)}, coverage.ceilingReached to ${ceilingReached}${suggestedNextActionClause}.\n\nArea maps (JSON):\n${JSON.stringify(raw, null, 2)}`,
   { label: 'synthesize', phase: 'Synthesize', model: 'fable', effort: 'xhigh', schema: REPORT_SCHEMA }
 )
--- a/.claude/workflows/tm-review-changes.js
+++ b/.claude/workflows/tm-review-changes.js
@@ -13,8 +13,9 @@ export const meta = {
 // one Fable critic. It cannot become the 100-agent fan-out that an unpinned
 // session-model review produces. Models and effort are pinned per
 // stage, so the session model and effort never leak into the workers; the
-// single critic runs Fable 5 at xhigh effort (flip its model pin to 'opus'
-// under the Opus 5 fallback, see team-guide).
+// single critic runs Fable 5 at xhigh effort and auto-retries once on opus
+// at the same effort if Fable returns nothing (criticWithFallback below;
+// see team-guide for the lead-level fallback, which is still manual).
 //
 // Invoke with an optional base ref:
 //   Workflow({ name: 'tm-review-changes', args: { base: 'origin/main' } })
@@ -26,6 +27,33 @@ function safeRef(value, fallback) {
   return typeof value === 'string' && /^[\w.~^\/\-]+$/.test(value) && !value.includes('..') ? value : fallback
 }
 const base = safeRef(args && args.base, 'origin/main')
+
+// Duplicated byte-for-byte across the three tm- workflows that run a Fable
+// critic (the workflow runtime has no shared imports); keep this copy in
+// sync, see helpers.test.mjs.
+async function criticWithFallback(prompt, opts) {
+  const first = await agent(prompt, opts)
+  // agent() returns null both when the model errors out after retries (for
+  // example Fable 5 quota exhaustion) and when the user skips the dispatch
+  // mid-run. Nothing in this script can tell those two cases apart, so there
+  // is no heuristic to invent here. The ruling is on which failure mode is
+  // worse: retrying a deliberate skip costs one extra skip prompt (the retry
+  // is itself an agent() call, so skipping again just returns null and the
+  // run ends below with a clear error), while not retrying a quota death
+  // costs the entire unattended run. Retry unconditionally.
+  if (first) return first
+  log(
+    `${opts.label}: the ${opts.model} critic returned nothing. This could be quota exhaustion or a manual skip; retrying once on opus at the same effort. Skipping again will end the run.`
+  )
+  const second = await agent(
+    `${prompt}\n\nNOTE: this is a retry on the opus fallback after the ${opts.model} critic returned nothing (quota or a skip). If you write a report file, record in it that this fallback produced it.`,
+    { ...opts, model: 'opus' }
+  )
+  if (!second) throw new Error(`${opts.label}: both the ${opts.model} critic and the opus fallback returned nothing.`)
+  log(`${opts.label}: this critique was produced by the opus fallback, not ${opts.model}.`)
+  second.modelFallback = `${opts.model} -> opus`
+  return second
+}
 
 // This list is diff-scoped and intentionally longer than tm-review-codebase's
 // per-area dimension list: docs and perf need diff context (what changed, and
@@ -141,7 +169,7 @@ const coverageNote = dropped.length
   : ''
 
 phase('Consolidate')
-const report = await agent(
+const report = await criticWithFallback(
   `You are the senior reviewer. ${covered.length} parallel reviewers produced the raw findings below.${coverageNote} ${diffHint}\n\nFor each raw finding: verify it against the actual diff, drop false positives and anything out of scope, merge duplicates, and set a final severity. You may add a finding only if it is a clear must-fix the reviewers missed. Only must-fix findings block: verdict is changes-requested if any remain, approve otherwise. Record every dropped finding under dismissed with the reason.\n\nRaw findings (JSON):\n${JSON.stringify(raw, null, 2)}`,
   { label: 'consolidate', phase: 'Consolidate', model: 'fable', effort: 'xhigh', schema: REPORT_SCHEMA }
 )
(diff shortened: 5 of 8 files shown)


sv-tmueller:
## Tester verdict, round 1 — issue #268: PASS

Commit under test: `c3efac4790e45323c0d7bc2ef2a92920d0611fe3`. **No findings.**

### Scope

`git show --name-only c3efac4` lists only `.github/workflows/ci.yml` — 38 insertions, 0 deletions, purely additive. The template and `ci-template-policy.test.mjs` are both untouched (`git diff` against each is empty). `npm test` reports exactly **65 pass, 0 fail**, matching the expected unchanged count, which is the proof neither policy-lock suite was disturbed.

### The four highest-risk checks

| Check | Result |
|---|---|
| YAML validity | Parses cleanly under PyYAML |
| Node pin | `actions/setup-node@v4` with `node-version: '22'`, plus a comment explaining why `cache: 'npm'` is absent |
| No cache, no install | Both confirmed absent — correct, since there is no lockfile and both would fail |
| `paths-ignore` and `types:` | Both absent, with an inline comment recording why |

The tester recomputed the markdown fraction rather than trusting the comment: **47/61 = 77.05%**, matching the stated 77%.

Triggers are `pull_request` (unrestricted, no `types:`, no path filters) plus `push` limited to `main`. No stray filter that would silently skip runs.

### One parsing note, not a defect

PyYAML resolves the unquoted `on:` key to boolean `true` rather than the string `"on"`, a YAML 1.1 quirk. Every GitHub Actions workflow in existence has this shape and GitHub's own parser special-cases it. Flagged so nobody "fixes" it later.

### Inherited risk, correctly judged out of scope

`node --test` with a glob matching nothing exits **0** with `# tests 0`, so a green CI run does not by itself prove tests ran. This is a pre-existing property of the repo's own `npm test` script, and fixing it requires an npm script change #268's non-goals forbid. Today the glob matches all 65 tests. **Real, but inherited rather than introduced.**

### Untested claim

Whether the workflow actually runs green on GitHub Actions cannot be verified from a checkout — there is no CI history on this branch and the first real execution is this PR. The tester said the file is well-formed and should run, and explicitly declined to claim it works.

**That claim is about to be settled by this PR itself.** The workflow triggers on the pull request that adds it, so the run appearing on #270 is the end-to-end verification no agent could perform. Watch that it reports 65 tests, not 0.

---

sv-tmueller:
## CI ran on its own PR and the untested claim is now settled

The tester explicitly declined to claim the workflow works, because no checkout can verify that. This PR settles it: the workflow triggered on the pull request that adds it, and completed in **8 seconds**.

More importantly, the log resolves the specific hazard flagged as "real but inherited" — that `node --test` with a glob matching nothing exits 0 with `# tests 0`, making a green run meaningless:

```
# tests 65
# suites 8
# pass 65
# fail 0
```

**65 tests actually executed on the runner.** Not a false green. The Node 22 pin does what it was pinned for: the quoted glob expanded correctly in GitHub's environment, which was the whole reason the sub-plan called that pin load-bearing rather than ceremonial.

That is the end-to-end verification the tester correctly said it could not perform, and it came from the artifact verifying itself.

## One thing only the live run could surface

The log carries a deprecation warning no local check could have produced:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/checkout@v4`, `actions/setup-node@v4`

This is about the **action runtime** — the Node that executes the actions' own JavaScript — not the Node 22 this workflow pins for `npm test`. Those are independent, and the test step is unaffected.

**Not acted on, deliberately.** It is a warning rather than a failure, the run is green, and bumping both actions to `@v5` would change a file that has already been tested and is mid-review. Recorded as a follow-up: `bump actions/checkout and actions/setup-node to v5` (`size:S`). It will keep warning until someone does, and will eventually break when GitHub drops the Node 20 shim.

Worth noting the shape of this: the repo's very first CI run immediately produced maintenance information that four rounds of local verification across two batches could not. That is the argument for the package, made by the package.

---

sv-tmueller:
## Reviewer verdict, round 1 — issue #268: APPROVE

Commit `c3efac4` reviewed in isolation. **Pass 1 clean** — all five criteria met, nothing extra beyond a two-line `permissions:` block. **Pass 2 sound, no must-fix.**

### The reviewer found a better reason than the sub-plan had

On dropping `paths-ignore`, the sub-plan argued 77% markdown and a sub-second suite. The reviewer found the decisive argument neither of us had:

> The markdown in this repo is not inert prose, it is test input. `effort-policy.test.mjs` reads `.claude/agents/*.md` frontmatter and asserts each seat's model and effort pins. Under the template's `paths-ignore: ['**.md']`, a PR editing only `.claude/agents/architect.md` — precisely the change that breaks the suite — would skip CI entirely.

Four of the last five merges to `main` were markdown-only. Keeping `paths-ignore` would have made the gate blind to the most likely breakage class in this repo.

### Node pin verified empirically, not argued

Tested against three runtimes rather than reasoned about:

| Runtime | Result |
|---|---|
| Node 20.20.2 | `Could not find '.../*.test.mjs'`, exit 1 |
| Node 21.7.3 | expands the glob, 65 tests |
| Node 22.22.2 | expands the glob, 65 tests |

So an unpinned runner defaulting to Node 20 would hard-fail. A matrix would be waste; `engines`/`.nvmrc` would be scope creep against the issue.

Also confirmed: `timeout-minutes: 5` gives roughly 7x headroom against a cold `setup-node`, and nothing here contradicts the team guide's CI cost policy (checked line by line — the e2e/draft bullet is vacuous with no e2e, and drafts still get the unit run because omitting `types:` keeps the default triggers).

### Should-fix, whose required fix is "not here"

**1. The zero-tests hazard is confirmed** — `node --test "nope/*.test.mjs"` exits 0 with `# tests 0`. If the test files are ever renamed or moved, this gate goes green having run nothing. Every fix routes through `package.json`'s test script, which non-goal 1 forbids. **Follow-up issue, not this PR.**

Note this PR's own CI run settles the present case: the log shows `# tests 65`, so the glob matches today. The hazard is about future moves.

### Nits — two taken, three left for you

**Taking:** the `node-version: '22'` line carries no comment despite being the least obvious line in the file (a hard floor, not a preference — a maintainer could "modernize" it to `lts/*` and break the gate). And the `paths-ignore` comment gives the two weakest arguments while omitting the decisive one, and its 77% figure is already **47/62 = 75.8%** once this file lands, so it will drift.

**Leaving for you:**
- `permissions: contents: read` is defensible least-privilege hygiene, not in the criteria and not in the template. The reviewer is not asking for removal, but flags whether the **template** should gain the same block — a separate issue, since non-goal 4 forbids touching it here
- The gate is inert until "Unit tests" is a required status check on `main`, which `NEW-PROJECT-SETUP.md:76` calls for. Repo settings, human-only

---

sv-tmueller:
## Tester verdict, round 1 — issue #269: FAIL

Commit `328a8fe`. One must-fix, and it is a good one.

### Must-fix: the site index made itself stale on insertion

Every doc-internal line citation in the addendum is off by a constant **+89 lines**. The numbers were computed against the pre-addendum file; inserting the 89-line addendum shifts everything from old line 18 downward by 89.

| Addendum cites | Actually now at | What the cited line contains |
|---|---|---|
| L72-76 (section 3) | 161-165 | inside the addendum's own text |
| L240-242 | 329-331 | section 4 pricing math |
| L244-254 ("Flag prominently") | 333-343 | section 6 capability quote |
| L256-262 (alias table) | 345-351 | still section 6 |
| L303-306 (section 8 rec.) | 392-395 | section 6 vendor positioning |
| L308-311 | 397-400 | section 8 body |
| L396-397 (section 9 trigger) | 485-486 | lands one line off, by coincidence |
| L376-392 (confidence caps) | 465-481 | the caps block |

```
sed -n '240,242p' docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
sed -n '329,331p' docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
```

Only L10-13 resolves, because it sits above the insertion point.

**Why this is must-fix rather than a nit.** When I declined to relax the acceptance criteria to allow inline pointers, I ruled that the site index plus the neutralised trigger sentences were an adequate substitute. The tester's judgment call, which I asked for explicitly, is that the substitute now fails for precisely the reader it exists to serve:

> For a reader who jumps directly into section 8 or 9 and never scrolls up, the addendum offers no protection at the point of read. The reader's only path back is the site index, and the site index's line numbers actively misdirect for six of eight entries.

So the mitigation I relied on does not currently work. That is my call coming back to be corrected, not a developer error in judgment.

### Everything else passed, and was verified rather than assumed

- **The two-hunk structural diff** — a pure 89-line insertion at the right place, and a one-for-one replacement at old line 241. Lines 236-242 byte-diffed against the pre-change file: identical apart from the one word. No reflow
- **Chronology correct** — `git show --stat daa176e` confirms the doc and the fix landed in the same squash commit, avoiding the error in my own issue text
- **Commit references accurate** — `git merge-base --is-ancestor daa176e HEAD` exits 0, `089f754` exits 1, so the unreachable flag is true
- **The confidence-cap claim independently checked** — all six caps read, none cites the effort-policy test. Medium is untouched, verified rather than asserted
- **Every "must not" honoured** — no revised recommendation or confidence, no strengthened lead-seat move, no claim that section 6 or the alias question is settled, correct "permits" rather than "endorses", and no mention of the stale event count
- Sections 1-10 present, in order, untouched. No em dashes, no cliches. Commit message conventional

### Nit

The agent sandbox had no `gh` and no GitHub tools, so it verified against the criteria as given rather than the live issue. Corroborated by the diff and doc content.

Fix round dispatched.

---

sv-tmueller:
### #269 - fix round 1 complete, re-test round 2 dispatched

`5cc6646` fixes the must-fix from test round 1 (site-index line numbers computed against the pre-addendum file, landing ~89 lines short).

The developer reports that the naive "+89" arithmetic was not sufficient: its own edits to the site-index bullets and the confidence-check paragraph shifted everything below them again, so the final numbers were recomputed against the actually-edited file and re-verified with fresh `sed` checks after the edits landed. It reports nine citations corrected (eight site-index entries plus the confidence-cap citation, which appears twice), `npm test` at 65/65, and a diff confined to the addendum block.

One disclosed deviation: a stale `L244-254` mention at line 103, inside the "Section 7 line drift" paragraph, was left untouched on the grounds that it was not one of the eight site-index entries or the confidence-cap citation the must-fix named. That is now with the tester to assess, since on its face it is the same defect the round failed on. The lead will arbitrate if the tester and the developer disagree.

Round 2 verification is running: every citation enumerated and resolved independently rather than sampled, the stable anchors checked for verbatim match, and the diff outside the addendum confirmed to be the single wording fix.

Process note: the fix commit was pushed with `--force-with-lease`. It landed as a strict fast-forward on `a0b8289`, so no history was rewritten and nothing was lost, but a plain push was what the situation called for.

---

sv-tmueller:
### #267 - developer DONE, test round 1 dispatched

`6049fb1` adds `criticWithFallback` to the three `tm-` workflows: the Fable critic stage that returns `null` is retried once on Opus 5 at the same `xhigh` pin, with the switch surfaced in the run output.

Five files, +204/-16: the three workflows, `helpers.test.mjs`, and `.claude/team-guide.md`. Test count goes 65 -> 70.

The developer reports both deliberate-break checks caught their target (a one-byte change in one copy failing the byte-identical assertion; a mismatched retry effort failing the retry test on `'high' !== 'xhigh'`), and that the three critic opts lines still carry `label:` + `model: 'fable'` + `effort: 'xhigh'` on one line each, so `effort-policy.test.mjs` still parses them. It rebased cleanly onto `5cc6646` from sibling package #269 before pushing.

The tester is reproducing both breaks independently with different bytes and a different file, since a claim that a test can fail is worth less than a demonstration of it, and is checking the both-null failure path at each of the three call sites rather than only in the helper.

Process note: this push also used `--force-with-lease`. Verified as a strict fast-forward - `a0b8289` and `5cc6646` are both ancestors of `6049fb1` and the history is linear, so nothing was lost. Two of two developer pushes this batch reached for a force flag they did not need; worth a look at the developer agent's push instructions once the batch is closed.

---

sv-tmueller:
### #269 - test round 2: FAIL

Verified `5cc6646`. Two of twelve citations still resolve to the wrong prose.

The tester declined the developer's count of nine and enumerated the citations itself:

```
grep -noE '(L[0-9]+(-[0-9]+)?|line [0-9]+(-[0-9]+)?|lines [0-9]+-[0-9]+|:[0-9]+)' \
  docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md | awk -F: '$1<124'
```

That finds **12** in-document citations, not 9. Ten resolve correctly with verbatim anchors. Two do not:

1. **`L244-254`** (line 103, "Section 7 line drift") claims the "Flag prominently" block quoting `effort-policy.test.mjs`. `sed -n '244,254p'` returns section 4 cost-math prose. The real target is `L350-360`. Disclosed by the developer as an intentional deviation.
2. **`L256-262`** (line 117, the closing paragraph) claims the alias-table question. `sed -n '256,262p'` returns more section 4 cost math. The real target is `L362-368`. **Not disclosed**, and already broken at `328a8fe` - the round-1 fix corrected the *site-index* copy of this citation to `L362-368` and left the independent second mention untouched.

Everything else holds: the doc body is unchanged apart from the single `failure` -> `blocker` word, `npm test` is 65/65, and the tester independently confirmed the addendum's factual claims about `089f754` (`git show 089f754` shows the `opus: 'xhigh'` addition; `git merge-base --is-ancestor 089f754 HEAD` fails, matching the "not reachable from this branch" claim; `effort-policy.test.mjs:26` matches the "line 26, not line 21" claim). No figure was quietly revised. UNTESTED CLAIMS: none.

The scope argument behind deviation 1 is rejected; reasoning is logged on #266. Short version: `5cc6646`'s commit message claims every citation was recomputed and verified, which is not true of the commit, so this is an incomplete fix rather than a boundary. Fix round 2 dispatched for both citations, specified as a property over the whole addendum rather than as a list of entries - the narrow reading was available because round 1's instruction was written narrowly, which is my error.

Tester counter: 2 of 3.

---

sv-tmueller:
### #267 - test round 1: PASS

Verified `6049fb1`. No findings. Review dispatched.

What the tester established rather than accepted:

- **Both deliberate breaks reproduced independently**, with a different byte and a different file than the developer used. A double space in `tm-review-codebase.js:69` produced `AssertionError: criticWithFallback diverged between tm-review-changes.js and tm-review-codebase.js`; adding `effort: 'high'` to the retry opts produced `'high' !== 'xhigh'`. Tree confirmed clean with `git status --short` after each revert, suite back to 70/70.
- **Byte-identical confirmed by its own extraction** (`awk` per file, pairwise `diff`, exit 0 on all three pairs) rather than by trusting the test that was itself under test.
- **The both-null path throws rather than returning null**, so no call site can proceed with a missing critic. All three call sites are `const report = await criticWithFallback(...)` with nothing dereferencing `report` before the return. Message: `consolidate: both the fable critic and the opus fallback returned nothing.`
- **The effort-policy suite still sees exactly what it saw before.** The new retry line `{ ...opts, model: 'opus' }` carries no `label:`, so it falls outside that test's `label: && model:` filter and the one pre-existing critic line per file is still the only match. `effort-policy.test.mjs` passes 17/17 on its own.
- **The quota-death vs user-skip decision is a decision, not a description.** The comment rules on asymmetric cost: retrying a deliberate skip costs one extra skip prompt, not retrying a quota death costs the whole unattended run. Retry unconditionally.

`npm test`: `# tests 70 / # pass 70 / # fail 0`, up from 65.

**One untested claim, settled by the lead rather than passed to review.** The tester flagged that it could not confirm `log` is a real Workflow-runtime ambient global - the repo's own codebase map lists only `agent`, `parallel`, `phase`, `args`, and this is the first bare `log(...)` in these three files. Its concern was sharp: the `node:vm` tests stub `log`, so a `ReferenceError` would be invisible to them and would fire only on the retry path, exactly when it is least recoverable. It is verifiable from the Workflow tool contract, which the tester had no access to: `log(message: string): void` emits a progress message shown as a narrator line above the progress tree. It is a genuine ambient global. Acceptance criterion 2 holds and the risk does not exist.

Its second untested claim (that the acceptance criteria it was given match issue #267) is a limitation of the sandbox, not of the change; the criteria were passed through verbatim and I have checked them against the issue.

---

sv-tmueller:
### #269 - fix round 2 complete, re-test round 3 dispatched

`47373af` corrects the two citations test round 2 found: `L244-254` -> `L350-360` and `L256-262` -> `L362-368`.

The whole diff:

```
 docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)
```

Both are 8-character in-place swaps, so no line in the document shifts and the ten already-correct citations stay correct. The developer verified both original targets landed in section 4's cost math before locating the replacements, rather than taking the corrected numbers from the dispatch on trust, and re-ran the enumeration after editing to confirm nothing moved. Pushed plainly, no force flag.

Round 3 is the cap. The tester has been told to derive its own enumeration and to distrust every count in its own prompt, including the developer's, since what failed rounds 1 and 2 was checking a handed-over list rather than the document. It has also been told not to soften a real finding to avoid the park.

---

sv-tmueller:
### #267 - review round 1: CHANGES_REQUESTED

Pass 1 (spec) clean: all seven acceptance criteria met, all four non-goals respected. The reviewer verified byte-identity with its own `awk`/`md5` slice (all three copies hash `9d6811bc4172db857f4c1eb27fe0c8e1`, 23 lines), confirmed `npm test` 70/70 in a throwaway worktree, and checked that only three `tm-` workflows exist so the team-guide's "every `tm-` workflow" is accurate. It also confirmed the loader changes weaken nothing: `(?:async\s+)?` is an optional group so `safeRef`/`parseArgs` still slice byte-for-byte, and `sandbox = {}` reproduces the old `createContext({})` exactly. The `.claude/team-guide.md` edit is not drive-by - the guide told a human to flip the workflow critic pins by hand and the code now makes that unnecessary.

Pass 2 findings:

**1. (must-fix) `helpers.test.mjs:487` - the wiring test cannot fail.**
`assert.ok(src.includes('criticWithFallback('))` is satisfied by the function declaration itself; each workflow contains exactly two occurrences, the declaration and the call site. Revert any call site to `await agent(` and the assertion still passes. This is the only test standing behind acceptance criterion 1's "in all three workflows" - the byte-identity test proves the helper exists, not that anything calls it. Fix: assert on `'await criticWithFallback('` (the declaration has no `await`) or on a count of at least 2, and verify the break by temporarily restoring `await agent(` at one call site.

Neither deliberate break run in this package touched this assertion; both targeted byte-identity. A suite can be verified breakable everywhere you thought to look and still contain an assertion that cannot fail.

**2. (should-fix) mutating the runtime's returned object**, at `tm-review-changes.js:54`, `tm-review-codebase.js:77`, `tm-map-codebase.js:84`. `second.modelFallback = ...` is strict-mode code; a frozen or proxied result throws a TypeError on exactly the fallback path this feature exists to protect, and the `node:vm` stub returns a plain object so the test cannot see it. Fix: return a new object rather than mutating.

**3. (should-fix) `effort-policy.test.mjs:72-74` documents a filter exception list this commit made incomplete.** The new spread opts carry `model:` with no `label:` and no literal `effort:` - a second exclusion the comment does not mention. The pin is genuinely preserved by the spread, so this is not a policy violation, but the comment now reads as exhaustive when it is not.

**4. (should-fix) `.claude/team-guide.md:199-201`** - the inserted parenthetical strands `.claude/workflows/` so it points at nothing, where it previously pointed at the pin list.

Findings 1 through 4 are in the fix round; reasoning for pulling 2-4 in despite the should-fix routing is logged on #266. Short version: each is a defect this commit introduced, in a file it already touches, and finding 2 sits on the recovery path itself.

**Held for your review, not fixed:**

- **5. (nit)** the helper's `opts` parameter shadows a module-level `const opts` in two of the three files. Harmless today since the helper only reads its parameter, but byte-identity fixes the name in all three copies, so a later edit reaching for the module-level `opts` would silently get the stage options.
- **6. (nit)** the quota-vs-skip comment earns its place but runs about two lines long ("so there is no heuristic to invent here" restates the previous sentence). Trimming pays three times, since the block is triplicated.

Reviewer counter: 1 of 3.

---

sv-tmueller:
### #269 - test round 3: PASS

Verified `47373af`. No findings. Review dispatched.

The tester was told to distrust every count in its own prompt, since checking a handed-over list is what let rounds 1 and 2 through. What it did with that:

- **Derived its own enumeration** and confirmed the addendum block boundary itself (`## Addendum` at line 18, `## 1. Question and scope` at line 124) rather than accepting line 124 from the prompt.
- **Swept for citation formats the regex could miss** - a broader raw-number extraction across the whole block. Every extra numeric token resolved to a date, a section number, an issue number, or a digit fragment of `daa176e`/`089f754`. No citation escaped enumeration.
- **Resolved all 12 doc-internal citations** with `sed -n`; every quoted anchor phrase appears verbatim at its cited location.
- **Verified the no-shift claim** rather than accepting it: both replacements are 8 characters for 8, confirmed with `wc -c`, no other line touched, no re-wrap.
- **Checked both external references**, which is subtler than it looks: the doc deliberately asserts "line 26, not line 21", so both numbers had to be right. `EFFORT_BY_MODEL` sits at line 21 with two keys pre-merge and line 26 with three keys post-merge. Both accurate.
- **Confirmed the squash**: `089f754` is genuinely unreachable from this branch, and its version of `effort-policy.test.mjs` is byte-identical to `daa176e`'s, which is what makes the addendum's account of the chronology correct rather than convenient.

Whole-package integrity holds: two hunks total, the addendum insertion plus the single `failure` -> `blocker` word. No figure, recommendation, or confidence level touched; section 8 still reads "Confidence: medium". `npm test` 70/70.

**One untested claim settled by the lead.** The tester could not reach GitHub to check whether PR #265 is the durable reference the addendum claims. It is merged, and its head SHA is exactly `089f754` - so the reference the addendum offers in place of the unreachable commit is the PR whose head commit is that commit. Stronger corroboration than the tester expected to be available.

**One forwarded to review rather than decided.** Whether naming `089f754` while pointing to `daa176e` and PR #265 as the runnable references satisfies the criterion's "and links it". That is a pass-1 spec question and it is the reviewer's call, not mine.

Tester counter closed at 3 of 3, with a PASS on the last available round.

---

sv-tmueller:
### #267 - review fix round 1 complete, re-test round 2 dispatched

`41ab06f`, six files, +34/-14. Plain push, no force flag.

- **Finding 1 (must-fix):** the wiring assertion is now `src.includes('await criticWithFallback(')`. The developer chose this over an occurrence count because the declaration line carries no `await`, so it cannot satisfy the string. Break-verified by restoring `await agent(` at a call site: `# pass 69 / # fail 1`, failing on `tm-review-changes.js does not call criticWithFallback(`.
- **Finding 2:** all three copies now `return { ...second, modelFallback: ... }` rather than assigning onto the runtime's object. Byte-identity confirmed by the developer's own `awk` extraction and pairwise `diff`. The existing tests assert on the value, not object identity, so they needed no change.
- **Finding 3:** the `effort-policy.test.mjs` comment now names the retry-opts line as a second deliberate exclusion and states why it is safe (effort inherited via spread from the caller line the test already checks).
- **Finding 4:** the team-guide sentence is split in two, with `.claude/workflows/` back where it points at something, plus a clause noting a long outage still burns one doomed Fable dispatch per run before the retry fires.

The two held-back nits (the `opts` shadowing, the comment length) are reported untouched.

Round 2 verification is running. The tester is reproducing the break on a different file than the developer used, and has been asked whether the new assertion is itself fully sound - whether anything other than a real call site could produce `await criticWithFallback(`, and whether the occurrence-count alternative would have been stronger. It has also been asked to check the team-guide's "one doomed dispatch" claim against the code, since one per run, per workflow, and per stage are different statements.

The standing instruction for this round is to find a second untestable assertion: for each new assertion, what would have to be true for it to pass while the feature is broken. Between them, the round-1 tester and the reviewer found one such assertion after both had already signed off on the surrounding suite.

Tester counter: 2. Reviewer counter: 1.

---

sv-tmueller:
### #269 - review round 1: APPROVE

Pass 1 clean. All three commits touch one file, the diff is two hunks (the addendum insertion and the one-word `failure` -> `blocker` change), and no figure, confidence level, body prose, or `.claude/team-guide.md` was touched. `npm test` 70/70.

**The forwarded spec question is ruled satisfied.** The reviewer confirmed `089f754` is unreachable from this branch while `daa176e` is the PR #265 squash carrying both the doc and the test change. Combined with #265's head SHA being `089f754`, the addendum names the commit the criterion names and links it through the PR that carries it. Citing an unreachable SHA as the runnable reference would have been the worse outcome.

It also noted something neither the issue nor I caught: **the addendum corrects the issue's own premise.** #269 says the blocker was cleared "a few hours after the doc was written"; the fix actually shipped in the same PR as the doc. The addendum says so rather than following the brief.

Six non-blocking findings, all yours:

**1. (should-fix) Three of eight anchor quotes span a hard line-wrap**, at doc lines 50, 55, 62, so the literal `grep` the addendum tells the reader to run at line 48 returns nothing for them. "the lead-only move is not" wraps across L12/13, "blocked by this repo's own test today" across L356/357, and "the effort-policy test gains an `opus` tier (a follow-up issue, not this package)" across L502/503. The other five match literally. This is the fallback added in round 1 specifically to survive drift, so it is the finding I would fix first. Shortening each quote to a fragment that fits one source line is enough.

**2. (should-fix) A mislabelled cross-reference at line 76.** "(section 8's second leg, below)" points at the effort-policy test, but section 8 names the test first and capability second, as does the addendum's own enumeration. A reader who counts finds section 8's second item is "The floor claim", about the lead move. It sits in the paragraph doing the load-bearing "what does not change" work.

**3. (should-fix) The addendum says its central conclusion three times** and repeats "confidence unchanged" and the line-21-vs-26 drift twice each. The reviewer's estimate: about 105 lines that would say the same thing in 60. Its verdict on whether the addendum soft-pedals the capability argument was the opposite of the worry - it over-insures.

**4. (should-fix, and it needs your decision rather than a fix round) Nothing points back.** The index points forward from the addendum to the stale sites, but a reader landing directly on section 8 still reads "leave architect, reviewer, and the three workflow critic stages on Fable 5 until the effort-policy test gains an `opus` tier" with no marker that the condition is now met. **That is the exact failure #269 was filed to prevent.** The developer could not fix it: criterion 3 and the "does not rewrite the body" non-goal forbid touching that prose. Either relax criterion 3 to allow a one-line "(see the 2026-07-25 addendum)" marker at sections 8 and 9, or accept the residual risk on record. I am not deciding this one - it is a change to the package's acceptance criteria, which is a parking-gate matter, and the package is otherwise complete.

**5. (nit, no action available)** `5cc6646`'s commit message claims it "recomputed every citation against the file as delivered and verified each with `sed -n`", which was false when written. The reviewer judged `47373af`'s message to correct the record adequately - it names both misses with their wrong and right targets and reports the full 12-citation enumeration - and recommended a sentence in the PR description over a force-push. That sentence will go in the PR body.

**6. (nit)** "named and neutralised" at line 69: British -ise against the document's own "organization", and it overclaims slightly, since the addendum cannot neutralise sentences it is forbidden to edit. "named and answered" was suggested.

**On the design question I asked.** The reviewer's answer: the line numbers are not the durable part and never will be, but they are not the liability either. The section headings in each index entry already carry the load unaided. Keep the numbers as a convenience the addendum openly labels as perishable. The weakness is finding 1 - the quoted anchors, added precisely to be self-correcting, are not all greppable.

Package #269 is complete: tester PASS, reviewer APPROVE. PR readiness waits on #267.

---

sv-tmueller:
### #267 - test round 2: PASS, with two findings that go back to the reviewer

Verified `41ab06f`. The must-fix break reproduced on a different file than the developer used: reverting `tm-review-codebase.js:277` to `await agent(` gives `# pass 69 / # fail 1`. The assertion can now fail.

The tester was asked to hunt for a second untestable assertion, on the grounds that the round-1 tester and the reviewer between them found one only after both had signed off on the suite. It found two.

**1. The hardened wiring assertion is still defeatable.** It is plain substring matching over raw source, so a decoy comment carrying the exact literal `await criticWithFallback(`, placed above a call site reverted to `await agent(`, leaves the suite fully green at 70/70. Reproduced on `tm-map-codebase.js`. It also tested the occurrence-count alternative the review offered and reports it is not stronger - the decoy comment counts toward the total too. It rates this non-blocking and distinguishes it from the round-1 bug, which was unconditionally blind because every file's own declaration always contained the bare substring; this one needs a deliberately crafted comment.

**2. The should-fix from review round 1 is behaviorally untested.** It reverted all three files to the mutation form, byte-identical to each other, and `npm test` stayed at 70/70. The `node:vm` stub always returns a plain mutable object, so the frozen/sealed/proxied case the fix defends against is never exercised. Reverting the fix is invisible to the suite.

That second one lands on a decision I made earlier in this batch. I pulled that should-fix into the fix round arguing it "sits on the recovery path this feature exists to protect". The fix is right, but the suite cannot tell it from the code it replaced, so the protection rests on reading the source rather than on anything enforced. Freezing the stub's return in the retry test would close that for a couple of lines. Whether it is in scope is the reviewer's call, not mine: the criterion says a test covers the retry path, not that a test covers the object-identity contract.

Everything else verified: the effort-policy filter's matched-line set is unchanged (only that file's comment was touched, not its logic); effort genuinely inherited via spread from a checked caller line; all three call sites wired; both-null still throws identically; byte-identity confirmed by the tester's own `awk` + `md5sum` rather than the repo's own test; clean tree after each of four temporary edits. On the team-guide's "one doomed dispatch per run" claim, it confirmed each workflow has exactly one call site, so per-run, per-workflow and per-stage coincide and the wording is accurate.

Re-review dispatched with both findings and the untested claim. Tester counter: 2. Reviewer counter: 1 of 3, so a fix round is available if the reviewer judges one necessary.

---

sv-tmueller:
### #267 - review round 2: APPROVE, plus one more fix round by choice

Pass 1 clean on `41ab06f`: exactly the six files, each round-1 finding has a corresponding edit, nothing rides along, conventional commit message, no em dashes or cliche vocabulary. The reviewer ruled the "one doomed Fable dispatch" clause not creep, since it sits inside the paragraph finding 4 required editing anyway and corrects a misreading the old text invited.

It verified the must-fix is genuinely repaired rather than merely changed: reverting `tm-map-codebase.js`'s call site with no decoy gives `not ok 5 - every workflow file wires the critic call through criticWithFallback`. It reproduced both tester findings, and tested the proposed remedy for the second: `Object.freeze({ verdict: 'approve' })` in the retry stub fails the mutation form and passes the shipped form.

**The package is approved and shippable. I am spending one more round on two findings anyway.**

**Finding 1 is my error.** The new comment states the mechanism as "workflow scripts are ES modules (strict mode)". The reviewer checked it: these files carry `export const meta` at line 1 and a bare top-level `return report` near the end, and top-level `return` is a SyntaxError in a real ES module. I confirmed it directly - `tm-review-changes.js:1` is `export const meta = {`, `:149` is `return report`. So the runtime is not evaluating these as ES modules and their strictness is established nowhere here.

That claim came from my round-1 fix instruction verbatim: "Workflow scripts are ES modules, so this is strict-mode code." The developer had no reason to doubt it and it is now in three files. The underlying advice survives on the reason the reviewer supplies instead: **mutating a frozen return silently drops the `modelFallback` marker even in sloppy mode**, which does not depend on strictness at all.

**Finding 2 is the one-word lock**, and it corrects something I wrote on this PR earlier. I justified pulling that hardening into the fix round on the grounds it "sits on the recovery path this feature exists to protect". Nothing enforced it: reverting all three files to the mutation form left the suite at 70/70. Freezing the stub's return makes the suite enforce what the source merely stated.

The reviewer argued both down to should-fix carefully - its own round-1 finding was a hardening rather than a shipped bug, so a test lock cannot outrank it, and the byte-identity test already catches a single-file revert, leaving only a coordinated three-file revert invisible. That reasoning is sound. I am fixing them anyway because one is a false statement about how this repo's code runs that I introduced, and the other is one word with a verified effect.

**Held for your review, not fixed:** the decoy-comment weakness in the wiring assertion (a full fix needs comment and string-literal stripping inside a zero-dependency `node:vm` suite, which is disproportionate); five comment lines for a one-string assertion; the failure message still reading `does not call criticWithFallback(` while the assertion checks `await criticWithFallback(`; a pointer that the effort-policy comment holds only because `EFFORT_BY_MODEL` maps both `fable` and `opus` to `xhigh`; and an orphan line left by the team-guide reflow.

If this round turns up anything not trivially fixable, `41ab06f` is already approved and ships instead. Counters after it: tester 3, reviewer 3, both at the cap.

### Separately: the CI failure on `41ab06f` is not a test failure

The run reports `failure` with **zero jobs created** and 76 seconds elapsed. I reproduced `npm test` locally at that exact commit on Node 22.22.2: 70/70, exit 0. And `ci.yml` is byte-identical between that run and the green run on its parent `47373af`, so nothing about the workflow changed under it. A run that creates no jobs failed before reaching any code. Re-running it produced no jobs either.

Worth recording before "Unit tests" becomes a required check: a zero-job startup failure is indistinguishable from a real failure in the checks UI, and it would block a merge on infrastructure noise. The next commit will produce a fresh run on a new SHA, which settles whether this was transient.

---

sv-tmueller:
### CI resolved: `41ab06f`'s failure was infrastructure, and the actions bump is now more urgent

**The zero-job failure was transient.** `9e011e3` ran green in 6 seconds, and the log confirms the run actually executed tests rather than hitting the zero-tests hazard:

```
1..9
# tests 70
# suites 9
# pass 70
# fail 0
```

That closes the question from the previous comment. `41ab06f` failed with no jobs created while `ci.yml` was byte-identical to the green run on its parent, and the same workflow on the next commit works. Nothing in the change caused it.

**But the deprecation warning has changed since #268's first run, and it is worse than recorded.** The first run warned that the actions targeted the deprecated Node 20. This run says:

```
##[warning]Node.js 20 is deprecated. The following actions target Node.js 20 but
are being forced to run on Node.js 24: actions/checkout@v4, actions/setup-node@v4.
```

They are no longer merely targeting a deprecated runtime; GitHub is now forcing them onto Node 24, a runtime they were not built against. That is a live compatibility risk rather than a housekeeping item, and it plausibly bears on the zero-job failure above, though I cannot show a causal link and will not claim one.

Two consequences for the follow-ups:

1. **Bump `actions/checkout` and `actions/setup-node` to v5.** No longer cosmetic.
2. **Do the bump before making "Unit tests" a required check.** A required check that can conclude `failure` with zero jobs would block merges on infrastructure noise, and the checks UI does not distinguish that from a real test failure. #268's own PR body argues that dropping `paths-ignore` is what makes the required-check flip safe; this is a second precondition, found by the CI observing itself.

Neither is filed. Both are in the batch report on #266 for you to accept or decline.

---

sv-tmueller:
### #267 - test round 3: PASS, no findings

Verified `9e011e3`. The mechanism behind the reworded comment is now confirmed twice over, independently of the developer.

The tester reverted all three workflows to the mutation form, keeping them byte-identical so the byte-identity test could not mask the result:

```
not ok 2 - first call returns null, retries on opus and succeeds
    + actual - expected
    + undefined
    - 'fable -> opus'
# tests 70, # pass 69, # fail 1
```

`undefined`, not a TypeError. A silent drop, which is what sloppy-mode assignment to a frozen object does. It then grepped `.claude/workflows/` for `'use strict'` and found **zero matches**. So sloppy mode holds both by the observed failure and by the absence of any strictness directive - and the clause I originally supplied ("ES modules, so strict mode") is now disproved from two directions rather than merely doubted.

It confirmed the export and top-level-return facts itself rather than taking them from the dispatch (line 1 `export const meta = {`; `return report` at 271, 180, 282), confirmed byte-identity by its own brace-depth extraction, and confirmed the commit touches exactly four files with `.claude/team-guide.md` and `effort-policy.test.mjs` untouched.

On whether the frozen stub creates collateral risk: that object is scoped to the one test and nothing else in the suite reuses or mutates it.

**One observation it raised and explicitly did not attribute to this diff.** If `second` ever legitimately carried its own `modelFallback`, a future rewrite with the property order flipped (`{ modelFallback: ..., ...second }`) would let `second`'s field win - satisfying this test, which never populates that field on the stub, while being wrong at runtime. The shipped order is correct and the risk predates this commit. Passed to the final review to judge, including the option of judging it not worth acting on.

Final review dispatched. Tester counter closed at 3, reviewer at 3 once it reports. If that review finds a genuine must-fix, the package ships on the already-approved `41ab06f` instead, losing these two improvements and nothing else.

---

PR #279, opened 2026-07-26
ci: bump checkout and setup-node actions to v5

sv-tmueller:
Closes #271

Part of batch #278.

## Why

The run log on `9e011e3` reported `actions/checkout@v4` and `actions/setup-node@v4` as being forced to run on Node.js 24, a runtime they were not built against. Earlier runs in the same batch warned only about the deprecated Node 20, so this escalated mid-batch. Moving both to the v5 major removes the forced-runtime path.

This is a precondition for making the unit-test check required: a required check that can fail for reasons unrelated to the code blocks merges on noise, and the checks UI does not distinguish that from a genuine failure.

## What changed

Two lines in `.github/workflows/ci.yml`: `actions/checkout@v4` -> `@v5` and `actions/setup-node@v4` -> `@v5`. Nothing else. `node-version: '22'` stays (the pin is load-bearing, since the test script's quoted glob is expanded by Node itself), and triggers, `timeout-minutes`, the concurrency group and `permissions` are untouched.

The v5 behavior changes in `setup-node` are inert here: the Node version is explicitly pinned and no `cache:` input is used, so v5's default-version and cache-resolution changes never engage.

`v5` was confirmed to exist as a current major by listing tags directly against each action's repository, not from documentation summaries: `actions/checkout` `refs/tags/v5` resolves to `v5.1.0`, `actions/setup-node` `refs/tags/v5` resolves to `v5.0.0`.

## Deliberately not changed

`.claude/skills/tm-new-project/templates/ci.yml` stays at v4. No test asserts action versions on it (`ci-template-policy.test.mjs` reads only that file and only checks structure), no test reads `.github/workflows/` at all, and the template belongs to the template-copy path that `docs/architecture/operating-model.md` section 3 records as retired with removal deferred to a later batch. Bumping a file slated for deletion would be scope creep on a `size:S` issue.

Making the check required in branch protection is a repository setting, out of scope here.

## Verification

Local `npm test` -> exit 0, 70 tests across 9 suites, all pass. That confirms nothing in the suite pins v4, but it does not verify the three acceptance criteria that live in the CI run log. Those are checked against this PR's own run: job success, no Node-version deprecation or forced-runtime warning, and a non-zero test count matching the local 70.

## Overlap to note when merging

Batch #278 also carries #272, which corrects the Node-floor comment a few lines below these two `uses:` lines in the same file. Whichever of the two PRs merges second may need a trivial conflict resolution in that hunk. The two changes do not otherwise interact.

---
[...]

Issue #278, opened 2026-07-26
Batch: test-gate-and-factual-corrections


sv-tmueller:
Approved contract for this batch. The proposal below is the signed-off scope; the decision log and the final report are in the comments. A dropped session resumes from this issue.

## Status: all four packages ready, awaiting human merge

| Package | PR | CI | Rounds |
|---|---|---|---|
| #271 CI actions to v5 | #279 | green on `d0fab64` | test PASS, review APPROVE, 0 fix rounds |
| #272 empty-test-run gate | #280 | green on `6fd0bf5` | 2 fix rounds, 3 test rounds, review APPROVE |
| #273 Fable-quota routing rule | #282 | green on `53faed5` | review APPROVE, 1 fix round, re-test PASS |
| #275 factual corrections | #281 | green on `f82c27f` | review CHANGES_REQUESTED, 1 fix round, re-review APPROVE |

Nothing parked. The full report is in the last comment; **#280 and #281 need hand-written squash messages**, see there.

## Approved proposal

Four ready-made packages handed over by the repo owner, already filed and sized. Refinement skipped per the advisor's ready-made-package rule; checked only for sizing and independence.

Verification done before proposing: `npm test` reports 70 tests / 9 suites, all passing (matches #272's stated count). `ci.yml` pins `actions/checkout@v4` and `actions/setup-node@v4` as #271 describes. `package.json` carries the single-line test script #272 targets. The codebase map lists the Workflow ambient globals without `log` at two places (`docs/architecture/2026-07-05-codebase-map.md:139` and `:190`), not one.

### 1. Bump the CI actions off the forced Node 24 runtime (#271) - size:S

Scope: move `actions/checkout` and `actions/setup-node` in `.github/workflows/ci.yml` from the v4 to the v5 major, so the job stops being force-migrated onto a Node runtime the actions were not built against. Verify from the actual PR run log that the forced-runtime warning is gone and the job still executes the real suite with a non-zero test count matching local `npm test` (70).

Acceptance criteria: as filed on #271.

Non-goals: `node-version: '22'` stays (load-bearing: Node expands the test glob); no change to triggers, `timeout-minutes`, concurrency, or `permissions`; does not make the check required in branch protection.

### 2. Make a green test run prove the tests actually ran (#272) - size:S

[...]

Issue #271, opened 2026-07-25
Bump the CI actions off the forced Node 24 runtime


sv-tmueller:
Part of batch #278

## What to build

The repo's CI job runs on actions that GitHub is no longer merely deprecating. The run log on `9e011e3` reports `actions/checkout@v4` and `actions/setup-node@v4` as "being forced to run on Node.js 24", a runtime they were not built against. Earlier runs in the same batch warned only that they targeted the deprecated Node 20, so this escalated mid-batch.

Move both to the v5 major, and confirm from the run log that the forced-runtime warning is gone and the job still executes the real test suite.

This is a precondition for making the unit-test check required. A required check that can fail for reasons unrelated to the code blocks merges on noise, and the checks UI does not distinguish that from a genuine failure.

## Acceptance criteria

- [ ] `actions/checkout` and `actions/setup-node` are pinned to the v5 major
- [ ] A CI run on the pull request completes successfully
- [ ] The run log carries no Node-version deprecation or forced-runtime warning
- [ ] The run log shows a non-zero test count matching the count from a local `npm test`

## Non-goals

- No change to the Node version the job installs; 22 stays, and the pin is load-bearing because the test script's quoted glob is expanded by Node itself
- No change to triggers, `timeout-minutes`, the concurrency group, or `permissions`
- Does not make the check required in branch protection; that is a settings action only the repo owner can take

sv-tmueller:
## Sub-plan (architect)

Dispatched by the /tm-advisor batch #278.

**Key question answered: the CI template does not need to change, and should not.**

- `.claude/workflows/__tests__/ci-template-policy.test.mjs` reads only the template file and asserts structure (the three job names, `timeout-minutes` per job, the concurrency group with `cancel-in-progress: true`, the draft-PR gate on build/e2e, `ready_for_review` in the trigger types). It asserts nothing about action versions, and no test in the suite reads `.github/workflows/` at all. A workflow-only bump keeps `npm test` green.
- The template (`.claude/skills/tm-new-project/templates/ci.yml`, three `checkout@v4` uses, no setup-node) belongs to the template-copy path that `docs/architecture/operating-model.md` section 3 records as retired, with physical removal deferred to the usability batch. Bumping a file slated for deletion is scope creep on a size:S issue, and this issue scopes the work to the repo's own CI job. Template stays at v4.

**Approach**

[...]

PR #280, opened 2026-07-26
test: fail npm test when the glob matches no test files

### LINKED ISSUE #272: Make a green test run prove the tests actually ran
Part of batch #278

## What to build

Running the test suite when it can find no test files currently reports success. `node --test` with a glob that matches nothing exits 0 and prints `# tests 0`, so a green CI check is not by itself evidence that anything was verified.

Make that case fail, with a message saying plainly that no tests were found.

Issue #268 documented this hazard and its non-goals forbade fixing it, because the fix needs a change to the test script rather than to the workflow. It matters more now than when it was recorded: the check is intended to become a required gate, and a gate that passes when it runs nothing is worse than no gate, because it looks like assurance.

The glob is expanded by Node rather than the shell, which is also why the Node version in CI is pinned. Whatever mechanism is chosen has to keep that property intact.

## Acceptance criteria

- [ ] A test run that matches no test files fails instead of exiting 0
- [ ] The failure message states that no tests were found
- [ ] The existing 70 tests still run and pass, with the count unchanged
- [ ] The behavior is demonstrated by temporarily pointing the pattern at a non-matching path, capturing the failure, and restoring it

## Non-goals

- No new dependency and no change of test framework
- No change to any existing test
- Does not restructure or move the tests directory
- Does not add coverage thresholds or any other new gate

sv-tmueller:
Closes #272

Part of batch #278.

## Why

`node --test` with a glob that matches nothing exits 0 and prints `# tests 0`. A green check was therefore not evidence that anything had been verified. That matters more now than when #268 recorded the hazard, because the check is intended to become a required gate, and a gate that passes when it runs nothing looks like assurance without being it.

## What changed

`scripts/test.mjs` (new, 20 lines, node builtins only) guards with `fs.globSync` and exits 1 with `no test files found matching <pattern>` when the pattern matches nothing. Otherwise it delegates to `node --test` via `spawnSync` with `stdio: 'inherit'` and propagates the child's exit code (`status ?? 1`, so signal death is a failure rather than a silent success). `package.json`'s test script becomes `node scripts/test.mjs`.

The pattern lives in exactly one place, a single `pattern` constant. That was the decisive argument against an inline `node -e` guard in `package.json`: two copies of the path inside a JSON string can drift, and a drifted guard checks one path while `--test` runs another, silently restoring the hazard.

Node-side glob expansion is preserved twice over. `spawnSync` is called without `shell`, so the pattern reaches `node --test` verbatim and never touches a shell at all.

Zero new dependencies, no framework change, no existing test modified, tests directory untouched, and `npm test` is still the entry point with the same output shape.

## Node floor comment corrected

Moving the glob into `scripts/test.mjs` made the workflow's floor comment wrong on both counts: it pointed at `package.json` for the glob, and it named 21.7.3 as sufficient. `fs.globSync` arrived in Node 22.0.0 (confirmed against Node's own documentation, which records it as "Added in: v22.0.0"), so the effective floor is now 22.0.0. Per CLAUDE.md the code wins and the doc is corrected in the same PR, so the comment is updated here.

## Argument order (fix round 1)

The first revision built `['--test', pattern, ...process.argv.slice(2)]`, placing forwarded arguments after the positional pattern. `node --test` requires its own flags to precede the pattern, so everything forwarded was silently discarded, and the original version of this PR body wrongly claimed `npm test -- <args>` parity was preserved. Corrected to `['--test', ...process.argv.slice(2), pattern]`.

For the record, this was not a regression introduced here. On `main` the script is `node --test "<pattern>"`, so `npm test -- --test-reporter=dot` expanded to `node --test "<pattern>" --test-reporter=dot` and was swallowed identically. Argument forwarding has never worked in this repo; the reorder makes it work for the first time. Plain `npm test` is unaffected, since with no extra arguments the spawn arguments are unchanged.

## Alternatives rejected

- **A runner flag.** None exists on the pinned Node major. `--test-force-exit` only forces exit after completion. Measured, not recalled.
- **A sentinel test asserting the suite is non-empty.** Structurally dead: if the glob is broken the sentinel does not run either.
- **A custom `--test-reporter`.** Reporters are not a reliable exit-code channel.
- **A `node:test` `run()` API wrapper.** Re-implements reporter selection, concurrency defaults and exit-code wiring, with parity risk. Guard-and-delegate keeps the child invocation close to what ran before.

`scripts/` is the repo's first source file outside `.claude/`. The runner is repo-dev machinery rather than plugin-shipped workflow machinery, so it does not belong under `.claude/workflows/`, and putting it inside the `__tests__/` directory it guards would be circular.

## Verification

The demonstration acceptance criterion 4 prescribes, run in order:

Green, after the wrapper:

```
$ npm test
# tests 70
# suites 9
# pass 70
# fail 0
EXIT:0
```

Pattern temporarily pointed at a non-matching path:

```
$ npm test
> test
> node scripts/test.mjs

no test files found matching does-not-exist/*.test.mjs
EXIT:1
```

Pattern restored, green again at 70 tests / 9 suites / exit 0. The committed `scripts/test.mjs` carries the working pattern; the broken one was never committed, confirmed by reading the file back out of the commit.

Additionally verified: signal death propagates (a `SIGKILL`ed child gives `status === null` and the `?? 1` yields exit 1); `npm test` behaves identically from the repo root and from a subdirectory, since npm sets cwd to the package root; and `fs.globSync` matches the same three files `node --test` then runs, so the guard and the delegated run agree for the pattern in use.

## Overlap to note when merging

Batch #278 also carries #271 (PR #279), which changes the two `uses:` lines two lines above the comment corrected here, in the same file and the same diff hunk. Whichever PR merges second needs a trivial conflict resolution. This branch is based on `origin/main` only and does not include #271's change; the two do not otherwise interact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,9 +34,9 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           # 22 is a floor, not a preference: npm test's glob (see
-          # package.json) relies on node --test expanding it, which
-          # Node 20.20.2 fails (exit 1, "could not find"); 21.7.3 and
-          # 22.22.2 both expand it and run the suite.
+          # scripts/test.mjs) relies on fs.globSync, added in Node 22.0.0,
+          # so the effective floor is 22.0.0, not the 21.7.3 that sufficed
+          # before the fail-on-empty-test-run wrapper existed.
           node-version: '22'
           # No cache: 'npm' here. There is no lockfile for setup-node's
           # cache-path resolution to key off, so it would fail.
--- a/package.json
+++ b/package.json
@@ -1,1 +1,1 @@
-{"private":true,"scripts":{"test":"node --test \".claude/workflows/__tests__/*.test.mjs\""}}
+{"private":true,"scripts":{"test":"node scripts/test.mjs"}}
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -1,0 +1,20 @@
+// Wrapper around `node --test` that fails when the glob matches nothing.
+// `node --test <glob>` on its own exits 0 with "# tests 0" for a
+// non-matching pattern, so a green run is not proof anything ran.
+import { globSync } from 'node:fs';
+import { spawnSync } from 'node:child_process';
+
+const pattern = '.claude/workflows/__tests__/*.test.mjs';
+
+const files = globSync(pattern);
+if (files.length === 0) {
+  console.error(`no test files found matching ${pattern}`);
+  process.exit(1);
+}
+
+const result = spawnSync(
+  process.execPath,
+  ['--test', pattern, ...process.argv.slice(2)],
+  { stdio: 'inherit' }
+);
+process.exit(result.status ?? 1);
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -14,7 +14,7 @@ if (files.length === 0) {
 
 const result = spawnSync(
   process.execPath,
-  ['--test', pattern, ...process.argv.slice(2)],
+  ['--test', ...process.argv.slice(2), pattern],
   { stdio: 'inherit' }
 );
 process.exit(result.status ?? 1);
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -12,9 +12,13 @@ if (files.length === 0) {
   process.exit(1);
 }
 
+// Arguments are deliberately not forwarded. `node --test` takes its flags
+// before the positional pattern, and a forwarded value-taking flag would
+// consume the pattern, leaving the runner to fall back to default file
+// discovery and report a false green.
 const result = spawnSync(
   process.execPath,
-  ['--test', ...process.argv.slice(2), pattern],
+  ['--test', pattern],
   { stdio: 'inherit' }
 );
 process.exit(result.status ?? 1);


sv-tmueller:
## Tester verdict, round 1: FAIL (1 finding)

Commit `bdd8a57`. All six acceptance criteria are independently met; the finding is against a claim this PR's body makes on its own initiative.

### Finding 1: argument forwarding is a silent no-op, contradicting the PR body

`scripts/test.mjs` builds `['--test', pattern, ...process.argv.slice(2)]`, placing forwarded args **after** the positional pattern. `node --test` requires its own flags to precede the pattern, so anything forwarded this way is swallowed with no error, no effect and no warning.

Reproduced from the repo root:

```
$ npm test -- --test-reporter=dot
# expected: dot-style reporter output
# observed: full default verbose TAP output, identical to plain npm test. exit 0, 70/9/70/0

$ npm test -- --test-name-pattern="nonexistent-test-name-xyz"
# expected: 0 tests match
# observed: all 70 tests run, unfiltered. exit 0

$ npm test -- some-bogus-extra-positional.mjs
# observed: silently ignored, still exactly 70 tests, exit 0
```

Root cause confirmed directly against `node --test`, bypassing the wrapper:

```
$ node --test '.claude/workflows/__tests__/*.test.mjs' --test-reporter=dot   # flag AFTER pattern -> ignored
$ node --test --test-reporter=dot '.claude/workflows/__tests__/*.test.mjs'   # flag BEFORE pattern -> honored
```

### Everything else passed

- **Criterion 6.** `npm test` from the repo root: exactly 70 tests / 9 suites / 0 fail / exit 0, output shape unchanged.
- **Criteria 1 and 2.** Reproduced without touching tracked files, by copying the wrapper to a scratch location and repointing `pattern`: `no test files found matching <path>`, exit 1. Working tree confirmed clean throughout.
- **Signal death.** A child killed by `SIGKILL` gives `status === null` and `signal === 'SIGKILL'`, and `status ?? 1` correctly yields exit 1. Verified with a standalone scratch script.
- **The `fs.globSync` floor claim, verified against the source rather than assumed.** Node's own docs record `fs.globSync(pattern[, options])` as "Added in: v22.0.0", so the corrected comment's floor is right, not merely plausible. No experimental warning is printed on a normal run.
- **Both floors named correctly.** 22.0.0 (`fs.globSync`) is the higher of the two constraints, above the older 21.7.3 (`node --test` glob expansion), and the comment does not mislead by naming only one.
- **cwd assumption.** `npm test` from the repo root and from `docs/` produce identical output; npm sets cwd to the package root regardless.
- **Glob parity.** `fs.globSync` matches the same three files `node --test` then runs, for the pattern actually used. No divergence.
- **Non-goals.** Diff is exactly the three claimed files, zero diff under `__tests__/`, no `dependencies`/`devDependencies` and still no lockfile, tests directory not moved, no new CI gate.
- **The `ci.yml` edit is comment-only.** Both `uses:` lines are byte-identical to `origin/main`, as intended; #271 changes them on its own branch.

### Untested claims

- The superseded comment's claim that Node 20.20.2 fails to expand the glob and 21.7.3 succeeds. Only v22.22.2 is available in the container, so the pre-22 behavior could not be exercised. Not part of this PR's own claim.
- Equivalence between `fs.globSync` and `node --test`'s glob matcher for pattern shapes other than the one in use (brace expansion, `**`, character classes). Both agree for the current pattern; other shapes are unverified.

### Lead decision on the fix

Sending this back for one fix round. The forwarded args move **before** the pattern, so the capability the code already reaches for actually works. Rationale and the reason this is not a regression are in the batch decision log on #278.

---

sv-tmueller:
## Tester verdict, round 2: FAIL (1 finding)

Commit `a6174ca`. The round-1 fix introduced a regression that reopens this issue's own hazard. Reproduced independently by the lead before acting.

### Finding 1: a forwarded value-taking flag swallows the pattern and produces a false green

Because forwarded args now precede the pattern, any forwarded flag that expects a value (`--test-concurrency`, `--test-name-pattern`, `--test-timeout`) consumes the pattern string as its own value, leaving `node --test` with zero positional patterns. Node falls back to its default test-file discovery, which matches any file named `test.mjs` and so picks up `scripts/test.mjs` itself. The wrapper is imported as a "test file" under a recursive `node:test` context and reports one trivial passing test.

```
$ npm test -- --test-concurrency
# Warning: node:test run() is being called recursively within a test file. skipping running files.
# Subtest: scripts/test.mjs
ok 1 - scripts/test.mjs
1..1
# tests 1
# suites 0
# pass 1
# fail 0
EXIT=0
```

Exit 0, one trivial test, and **none of the 70 real tests ran**. Same with `--test-name-pattern` and `--test-timeout` given no value. Reproduced directly without npm: `node --test --test-concurrency '.claude/workflows/__tests__/*.test.mjs'` gives identical output.

**This is a regression from the fix, not a pre-existing condition.** On the parent commit `bdd8a57`, where forwarded args were silently discarded, the identical invocation ran all 70 real tests. Both cases exit 0, so CI cannot tell a real pass from this false pass by exit code. That is precisely the "a green run is not proof anything ran" hazard this issue was opened to close, now reachable through a forwarded argument instead of an empty glob.

### Everything else confirmed

- Plain `npm test`: 70 / 9 / 0 fail, exit 0, unchanged.
- `npm test -- --test-reporter=dot` produces dot output; `npm test -- --test-name-pattern="nonexistent"` produces three trivial suite wrappers with empty bodies. Both are the flags genuinely taking effect. The name-pattern quirk is inherent to `node --test-name-pattern`, independent of argument order, and out of scope here since this issue is about empty *file* matches rather than empty *name* matches.
- Empty-match guard: still exit 1 with the right message, no regression.
- The committed pattern constant is the working one on both `bdd8a57` and `a6174ca`; the broken pattern was never committed.
- `git diff bdd8a57..HEAD` is exactly the one-line reorder. `git diff origin/main...HEAD --stat` is still the three files, with `ci.yml` comment-only and both `uses:` lines byte-identical to `origin/main`.
- A genuinely failing test still exits non-zero.
- A forwarded `--` separator adds positionals as a union rather than a filter, so all 70 still run. No hole there.

### Lead decision, reversing my round-1 call

My round-1 instruction was wrong, and the evidence now says so. I chose to reorder the arguments to make forwarding work; that traded a harmless no-op for a silent false-green, which is strictly worse and defeats the package's own purpose.

**Fix round 2: remove the argument forwarding entirely.** The spawn becomes `['--test', pattern]`, with a comment stating that arguments are deliberately not forwarded and why. Rationale in the batch decision log on #278.

Forwarding cannot be made safe here without parsing Node's own flag table to know which flags consume a value, which is far out of proportion for this issue and would be new machinery no acceptance criterion asks for. No acceptance criterion mentions argument forwarding at all; all six were met at round 1 and stay met without it.

---

sv-tmueller:
## Tester verdict, round 3: PASS

Commit `6fd0bf5`. No findings. All four acceptance criteria re-verified from scratch rather than carried forward, since the file changed twice since round 1.

- **Criteria 1 and 2.** A non-matching pattern exits 1 with `no test files found matching <pattern>`, reproduced via a scratch copy so no tracked file was touched. `git status` stayed clean throughout.
- **Criterion 3.** `npm test`: 70 tests / 9 suites / 70 pass / 0 fail, exit 0.
- **Criterion 4.** Demonstrated entirely from a scratch copy.

### The round-1 regression is gone

`npm test -- --test-concurrency=1`, `npm test -- --test-name-pattern ZZZ` and `npm test -- --test-timeout 5000` all run the real 70 tests, with no `ok 1 - scripts/test.mjs` and no single-test false green. Arguments are never read from `process.argv`, so injection is structurally impossible rather than merely guarded against.

**The comment's justification was verified rather than trusted.** The new comment claims a forwarded value-taking flag "would consume the pattern, leaving the runner to fall back to default file discovery and report a false green". The tester reproduced exactly that against the round-1 commit `a6174ca`: `node <a6174ca-wrapper> --test-name-pattern` produced `ok 1 - scripts/test.mjs`, `# tests 1`, exit 0. The comment describes a real mechanism, not a hypothetical.

### Other checks

- Exit-code propagation: a genuine failing test exits 1. Signal death (`status: null`, `signal: 'SIGTERM'`, also reproduced with a real killed child) yields exit 1 through `status ?? 1`.
- The working pattern is committed on all three commits (`bdd8a57`, `a6174ca`, `6fd0bf5`); the broken demonstration pattern was never committed at any point.
- Diff is exactly the three files. `ci.yml` comment-only with both `uses:` lines byte-identical to `origin/main`. `package.json` changed only its `test` script, no `dependencies`/`devDependencies`, still no lockfile. Zero diff under `__tests__/`.
- Running `npm test` from a subdirectory still works, since npm sets cwd to the package root.
- Guard-versus-runner divergence: no case found where `fs.globSync` and `node --test` disagree for the pattern in use. A directory named `weird.test.mjs` matched by the glob correctly fails rather than passing.

### One observation, judged out of scope

`NODE_OPTIONS='--test-name-pattern=ZZZ_NO_MATCH' npm test` exits 0 while most real subtests silently do not run (3 of 9 file-suites reported as trivially ok, no real assertions). That is a genuine route to a green run that verified little.

Out of scope here, and I agree with the tester's reasoning: it predates this PR (the bare `node --test <glob>` on `main` has identical exposure, so it is not a regression), it requires an operator to deliberately set an environment variable rather than something an innocent caller passes, and this issue's acceptance criteria concern the glob matching zero *files*, not env-driven *name* filtering after files are found. `fs.globSync` operates at file-discovery level and was never designed to see into `NODE_OPTIONS`.

Flagging it for the human because the issue's motivation is that the check is intended to become a required gate, and this is a remaining way that gate can be green while verifying little. If you want it closed, it needs its own issue: the natural fix is asserting on the run's reported test count rather than only on file discovery, which is a different mechanism from the one this package chose.

Dispatching the reviewer.

---

sv-tmueller:
## Reviewer verdict: APPROVE (2 should-fix, 1 nit)

All four criteria verified by reproduction rather than trusted, including re-reproducing the round-1 false green against the `a6174ca` wrapper to confirm the final comment's justification is accurate rather than plausible.

### The four judgment calls

- **Mechanism sufficiency.** The reviewer agrees the guard proves files were *found*, not that tests *ran*, and that the `NODE_OPTIONS` gap is real. It still holds that guarding at file-discovery level is a sufficient reading of this issue: the acceptance criteria are written entirely in terms of the glob matching zero files, the exposure predates this PR identically, it needs a deliberately set environment variable, and non-goal 4 forbids "any other new gate", which a minimum-test-count assertion would be. It adds one thing worth carrying: the count assertion should be filed **before** the check becomes a required gate, because the issue's gate motivation is only fully served at run level.
- **`ci.yml` comment correction belongs here.** The old comment became false twice over once the code moved: "see package.json" no longer held, and the floor rationale changed. The reviewer also checked the overlap concretely: #279 changes lines 33-34, this PR changes only lines 36-39, and the `uses:` lines are byte-identical to `main`. Different lines, clean textual merge. Better than I expected when I accepted the overlap; I had assumed a likely conflict.
- **Floor comment accuracy, verified against the source.** Node's latest-v22.x docs record `fs.globSync` as "Added in: v22.0.0", and the latest-v21.x `fs` docs contain zero occurrences of `globSync`, so no 21.x release suffices and 22.0.0 is genuinely binding over `node --test`'s 21.7.3. One immaterial note: `globSync` was experimental until 22.17.0, and CI's `'22'` resolves above that.
- **Branch history.** Leaving `a6174ca` (which introduced the hazard) and `6fd0bf5` (which removed it) as-is is acceptable. The repo squash-merges, so `main` receives one commit and the round-trip survives only on the branch, where it is the audit trail of a reversal. Rewriting to hide the mistake gains nothing and destroys that record.

### Finding 1, should-fix: CLAUDE.md's Repo layout omits `scripts/`. Declining, with reasoning.

The reviewer holds that `CLAUDE.md:47` now under-describes the repo, since this is the first source directory outside `.claude/` and `docs/`, and notes that nothing in the doc becomes false.

Not applying it, because the counter-argument was not weighed: that block is already selective rather than exhaustive. It documents the `docs/` subtree only, and omits `.claude/` (covered in `team-guide.md` instead), `.github/`, `e2e/`, and `package.json`. Adding `scripts/` while still omitting `.claude/` and `.github/` would make the block less coherent, not more. CLAUDE.md's "code wins, the doc is corrected" rule is about disagreement, and by the reviewer's own assessment there is none here, only incompleteness that predates this PR in several other ways.

If you want the layout block made exhaustive, that is a worthwhile one-issue cleanup covering all the omissions at once, rather than a single line bolted on here. Recorded for you to decide.

### Finding 2, should-fix: file a follow-up for the run-level assertion

The `NODE_OPTIONS='--test-name-pattern=ZZZ' npm test` route still exits 0 while most real subtests silently do not run. Out of scope here on all three grounds above, and the reviewer explicitly says do not change this PR.

I am not filing that issue unilaterally: the batch sign-off covered four packages, and adding a fifth issue to the repo is a new artifact outside the approved contract. It is in the batch report on #278 as a recommended follow-up, and I can file it on request. It matters most before the unit-test check becomes required in branch protection, which is what #271 was a precondition for.

### Nit 3: commit type, for the squash message

`bdd8a57` uses `test:` for a change that touches runner infrastructure, `package.json` and a CI comment; `fix:` matches the intent better. Moot on the branch.

**At merge time:** the squash message must describe the final state (glob guard, no argument forwarding) rather than concatenating the three commit bodies, which would describe a forwarding feature that does not exist.

### Pipeline complete

Last tester verdict PASS (round 3), reviewer APPROVE, two fix rounds. Marking this PR ready for review.

---

Issue #273, opened 2026-07-25
Make the Fable-quota Opus override a routing rule, not a remembered procedure


sv-tmueller:
Part of batch #278

## What to build

Closes the dispatchable remainder of #261. Issue #267 automated recovery for the workflow critic stages, which was the half that mattered most because workflow scripts run unattended. Lead-level dispatches still depend on the lead noticing a quota death and re-dispatching that one agent with a per-call Opus override.

That procedure works. It is documented in the Model policy as prose the lead has to remember and apply. The evidence that remembering is not enough is now substantial: across two consecutive batches the Fable 5 limit was hit seven times on four separate occasions, and in the most recent batch **every** judgment dispatch died on quota and needed the override applied by hand.

Make it an explicit routing rule in the dispatch pipelines, so a fresh lead session applies it without rediscovering it from policy prose, and so the model switch is recorded rather than silent.

The existing manual-fallback prose in the Model policy has to be reconciled with the new rule, not left beside it saying something different.

## Acceptance criteria

- [ ] The kickoff per-package pipeline's routing rules name a Fable-quota agent death and prescribe the per-call Opus override as the response
- [ ] The rule requires the switch to be logged as a decision, so a model change is never silent
- [ ] The advisor's in-scope-decision rules cover the same case and do not contradict the kickoff rule
- [ ] The Model policy's existing manual-fallback prose is reconciled with the new rule rather than left to conflict with it
- [ ] `npm test` passes

## Non-goals

- Fable 5 stays the default judgment model; nothing changes about which model is tried first
- No `effort:` pin changed and no agent frontmatter edited
- Does not address the plugin-cache-overwrite problem #261 raises, which needs its own design pass
- Does not add a batch-level "Fable is down" switch
- Does not introduce Sonnet as a judgment fallback

sv-tmueller:
## Sub-plan (architect)

Dispatched by the /tm-advisor batch #278. Read against `docs/architecture/operating-model.md`: no conflict, since the operating model delegates model policy to team-guide and pipeline mechanics to the kickoff skill, which is exactly where this change goes.

### Approach

Three docs-only edits, all using the same phrases ("per-call Opus override", "the Agent tool's `model` param") so grep finds them together. The kickoff routing rule is the single owner of the mechanics; the advisor and the Model policy reference it rather than restate it. The rule is described as the lead-level complement of `criticWithFallback` (#267), which already covers workflow critic stages, so the two cannot be confused or double-applied.

[...]

PR #281, opened 2026-07-26
fix: correct the factual defects batch #266's reviews found

### LINKED ISSUE #275: Correct the factual defects batch #266's reviews found
Part of batch #278

## What to build

Batch #266 shipped with several non-blocking findings recorded rather than fixed. This package takes only the ones that are factually wrong, and deliberately leaves the ones that are matters of taste.

**Anchors that defeat the instruction next to them.** The addendum in `docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md` tells the reader to `grep` for a quoted phrase when a line number has drifted. Three of its eight quoted anchors span a hard line-wrap in the body, so that `grep` returns nothing for exactly those three. The anchors exist to survive drift, and these do not.

**A cross-reference that miscounts.** The addendum calls the effort-policy test "section 8's second leg". Section 8 names the test first and capability second, and its second enumerated item is something else entirely. The sentence sits in the paragraph doing the load-bearing "what does not change" work.

**A failure message naming the wrong string.** The workflow wiring test asserts on one string and reports failure using a different one, so a future failure points at something the assertion no longer checks.

**A comment clause that argues against both options.** The `criticWithFallback` comment justifies returning a new object rather than mutating, and closes with a note that a proxied result could throw either way. That is true and it does not distinguish the two forms, since spreading a proxy runs its traps just as assigning to one does. It is residue from an earlier wording that also contained a claim since removed as false.

**A documented list that is missing an entry.** `docs/architecture/2026-07-05-codebase-map.md` enumerates the Workflow runtime's ambient globals as `agent`, `parallel`, `phase` and `args`. `log` is also one, and its absence had a real cost: a tester could not verify a correct `log()` call from inside the repo and had to raise it as an unverifiable risk.

## Acceptance criteria

- [ ] Every quoted anchor in the addendum's site index matches its target verbatim on a single source line, verified by running the `grep` the addendum itself recommends
- [ ] The "section 8's second leg" cross-reference describes what that section actually says
- [ ] The wiring test's failure message names the string the assertion checks
- [ ] The proxy clause is removed from the `criticWithFallback` comment, which stays byte-identical across all three workflows
- [ ] `log` is listed among the Workflow ambient globals in the codebase map
- [ ] `npm test` passes

## Non-goals

- Does not shorten the addendum or reduce its repetition. Rewriting a dated record after the fact is the error batch #262 declined to make, and that call belongs to the repo owner
- Does not rename the helper's `opts` parameter, which shadows a module-level binding in two of three files but is read-only inside the helper
- Does not trim the quota-versus-skip comment
- Does not attempt to make the wiring assertion resistant to a deliberately crafted decoy comment; closing that properly needs comment and string-literal stripping inside a zero-dependency test suite
- Does not change any figure, recommendation, or confidence level in the research doc

sv-tmueller:
Closes #275

Part of batch #278.

Five factual corrections from batch #266's non-blocking findings. Only the things that are wrong; the matters of taste are deliberately left alone.

## 1. Three grep anchors that defeated the instruction beside them

The addendum's site index tells the reader to `grep` for a quoted phrase when a line number has drifted. Three of its eight anchors quoted a phrase that spans a hard line-wrap in the body, so that `grep` could never find the target. Each is shortened to a substring that lives verbatim on one target line, and each entry keeps its line count so no `L` reference anywhere in the addendum drifts.

Shortening the quote (rather than re-flowing the body) is the only option available: the addendum's own rule is that "None of that prose is edited by this addendum", and the issue's non-goals forbid rewriting the dated record.

Verified by running the `grep` the addendum itself recommends, for all eight anchors:

```
anchor1 'the lead-only move is'                                                   -> 12, 50
anchor2 'the one that would fail the effort-policy test'                           -> 180
anchor3 'neither is blocked by the effort-policy blocker below'                    -> 347
anchor4 "blocked by this repo's own"                                              -> 356
anchor5 'the same `shared/models.md` alias table'                                  -> 362
anchor6 'until the effort-policy test gains an `opus` tier'                        -> 411
anchor7 'This is option (c), not (b)'                                              -> 61, 414
anchor8 'effort-policy test gains an `opus` tier (a follow-up issue, not this package)' -> 503
```

Every anchor reaches its target. The extra hits at L50 and L61 are index entries quoting themselves, which anchor 7 already did before this change.

**Meaning preserved alongside the grep fix (review fix round 1).** Anchor 1's entry carries a gloss, and on `main` the quote plus that gloss reconstructed the summary's elided predicate: `"the lead-only move is not" (blocked by the test)`. Removing `not` from the quote to fix the wrap left the gloss asserting the opposite of the target, so the gloss is now `(not blocked by the test)`. L12-13 reads "the lead-only move is / not." and L417 states "The lead-only move is not blocked by that test", so the corrected entry matches both. The anchor, the line count, and every grep result above are unchanged by that fix.

One precision note on the issue's own wording: it said the broken `grep` "returns nothing for exactly those three". Anchors 4 and 8 did return nothing, but anchor 1 returned exactly one line, which was the index entry quoting itself rather than the target. Same defect, same fix.

## 2. A cross-reference that miscounted

The addendum called the effort-policy test "section 8's second leg". Section 8 names the test first (L414) and the capability claim second (L415-416), and its second enumerated item is the floor claim. The addendum's own two-legs paragraph (L82-84) orders test first as well. Corrected to "first leg". One word, in the paragraph doing the load-bearing "what does not change" work.

## 3. A failure message naming a string the assertion does not check

`helpers.test.mjs:492` asserted on `'await criticWithFallback('` but reported failure as `does not call criticWithFallback(`. Per the test's own comment, the bare `criticWithFallback(` substring is present even in a file that has the defect, so the message pointed at something that would exist in the failing file. The message now names the checked string.

Per the issue's non-goals, the assertion itself is untouched: making it resistant to a deliberately crafted decoy comment needs comment and string-literal stripping inside a zero-dependency test suite, which is separate work.

## 4. A comment clause that argued against both options

The `criticWithFallback` comment justifies returning a new object rather than mutating, then closed with a note that a proxied result could throw either way. That is true but it does not distinguish the two forms, since spreading a proxy runs its traps just as assigning to one does. It was residue from an earlier wording that also carried a claim already removed as false.

Removed from all three workflow files in one commit, since the comment sits inside the function body and `helpers.test.mjs` asserts the helper is byte-identical across all three. Splitting the change would have failed that test. The remaining sentence reads complete after the cut, not clipped.

## 5. A documented list missing an entry

The codebase map enumerated the Workflow runtime's ambient globals as `agent`, `parallel`, `phase` and `args`. `log` is also one, and the omission had a real cost: a tester could not verify a correct `log()` call from inside the repo and had to raise it as an unverifiable risk. All three workflow scripts call `log()` free and nothing declares it.

Both enumerations are corrected, the dependency summary and the open-questions section, not just one. One fixed and one stale would be the same defect the issue is closing.

**What kind of edit this is, stated precisely.** The map is a dated generated artifact and the repo convention supersedes rather than edits older reports, so the category matters for the precedent. This is **a sign-off-sanctioned update of a snapshot that went stale, not the correction of an error the artifact had when written.** At the map's generating commit (`d06fd07`, 2026-07-05) no workflow script referenced `log` at all; `log()` entered with `9465e67` (#270, 2026-07-25). The map was accurate on its own date.

The in-place edit is still the right call: criterion 5 requires it, batch #278's sign-off authorized it, and CLAUDE.md's "code wins, the doc is corrected" rule covers the map's standing claim against current code, while regenerating an entire map to add one word would violate simplicity-first. Recording the category explicitly because in-place edits to dated artifacts stay a safe precedent only when they are either corrections of same-date errors or labelled sanctioned updates. Commit `787b9a2`'s body frames this as an original omission, which is wrong; the framing here and in the squash message is the accurate one.

## Verification

- All eight anchor greps resolve to their target lines, transcript above, re-run after the review fix round.
- `npm test` -> exit 0, 70 tests / 9 suites, run after every commit. This covers the byte-identity assertion (proving item 4 landed identically in all three files), the wiring test, and the effort-policy suite.
- `grep -rn 'proxied' .claude/workflows/` returns nothing.
- Both codebase-map enumerations now list `log`.
- The addendum is 548 lines on both `origin/main` and this branch, so no `L` reference drifted.

## Held to the non-goals

The addendum is not shortened and its repetition is not reduced; the helper's `opts` parameter is not renamed; the quota-versus-skip comment is not trimmed; the wiring assertion is not hardened; and no figure, recommendation, or confidence level in the research doc changed.

## Noted, deliberately not fixed

`docs/plans/37-review-codebase.md:37` also enumerates the ambient globals without `log`. Left alone: plan docs are point-in-time snapshots by repo convention with no standing-truth duty, and this issue's scope names the codebase map. A one-line follow-up if you want it corrected.

## For the squash message

Two framing corrections worth carrying over, neither worth a history rewrite: the `log` addition is a sanctioned update of a stale snapshot rather than an original omission (see section 5), and commit `354a78a` uses `fix:` for a comment-only change where `docs:` fits the taxonomy better (defensible, since the comment asserted a claim that did not hold).

### DIFF
--- a/.claude/workflows/__tests__/helpers.test.mjs
+++ b/.claude/workflows/__tests__/helpers.test.mjs
@@ -489,7 +489,7 @@ describe('criticWithFallback', () => {
     // string only matches an actual call.
     for (const file of FILES) {
       const src = readFileSync(join(workflowsDir, file), 'utf8')
-      assert.ok(src.includes('await criticWithFallback('), `${file} does not call criticWithFallback(`)
+      assert.ok(src.includes('await criticWithFallback('), `${file} does not contain 'await criticWithFallback('`)
     }
   })
 })
--- a/.claude/workflows/tm-map-codebase.js
+++ b/.claude/workflows/tm-map-codebase.js
@@ -84,7 +84,7 @@ async function criticWithFallback(prompt, opts) {
   // Return a new object rather than mutating `second`: the runtime's returned
   // object could be frozen or sealed, in which case mutating it would either
   // silently drop the `modelFallback` marker (sloppy mode) or throw a
-  // TypeError, and a proxied result could throw either way.
+  // TypeError.
   return { ...second, modelFallback: `${opts.model} -> opus` }
 }
 
--- a/.claude/workflows/tm-review-changes.js
+++ b/.claude/workflows/tm-review-changes.js
@@ -54,7 +54,7 @@ async function criticWithFallback(prompt, opts) {
   // Return a new object rather than mutating `second`: the runtime's returned
   // object could be frozen or sealed, in which case mutating it would either
   // silently drop the `modelFallback` marker (sloppy mode) or throw a
-  // TypeError, and a proxied result could throw either way.
+  // TypeError.
   return { ...second, modelFallback: `${opts.model} -> opus` }
 }
 
--- a/.claude/workflows/tm-review-codebase.js
+++ b/.claude/workflows/tm-review-codebase.js
@@ -77,7 +77,7 @@ async function criticWithFallback(prompt, opts) {
   // Return a new object rather than mutating `second`: the runtime's returned
   // object could be frozen or sealed, in which case mutating it would either
   // silently drop the `modelFallback` marker (sloppy mode) or throw a
-  // TypeError, and a proxied result could throw either way.
+  // TypeError.
   return { ...second, modelFallback: `${opts.model} -> opus` }
 }
 
--- a/docs/architecture/2026-07-05-codebase-map.md
+++ b/docs/architecture/2026-07-05-codebase-map.md
@@ -136,8 +136,8 @@ no dependency fields):
 - Claude Code harness: CLAUDE.md loading and @-imports, the plugin/marketplace
   system (reads marketplace.json and plugin.json), settings.json, the
   subagent/Task machinery that runs the role agents, and the Workflow tool
-  runtime that supplies the ambient `agent`, `parallel`, `phase`, and `args`
-  bindings the three scripts reference but never define.
+  runtime that supplies the ambient `agent`, `parallel`, `phase`, `log`, and
+  `args` bindings the three scripts reference but never define.
 - GitHub via the `gh` CLI: issues, PRs, labels, comments, and search are both
   the input source and the durable state store for tm-kickoff, tm-advisor,
   tm-to-issues, tm-new-project, and every role agent.
@@ -187,10 +187,10 @@ Cross-area dependencies:
 ## 5. Open questions
 
 - The Workflow runtime contract is assumed, not defined: `agent`, `parallel`,
-  `phase`, and `args` are ambient globals the three scripts call but nothing
-  in the repo declares. Their exact semantics (failure handling, null-padding,
-  schema enforcement) can only be confirmed against the Claude Code host, not
-  from source here.
+  `phase`, `log`, and `args` are ambient globals the three scripts call but
+  nothing in the repo declares. Their exact semantics (failure handling,
+  null-padding, schema enforcement) can only be confirmed against the Claude
+  Code host, not from source here.
 - operating-model.md defers the physical removal of NEW-PROJECT-SETUP.md and
   .claude/skills/tm-new-project/ to a later usability batch. The timing is a
   human decision, and the fate of templates/ci.yml, which
--- a/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
+++ b/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
@@ -47,7 +47,7 @@ blocker that is now cleared. Line numbers are current as of this commit; a
 document that edits itself drifts, so each entry also carries a section
 heading and a quoted phrase to `grep` for if the number is wrong again:
 
-- L10-13, the summary: "the lead-only move is" (blocked by the test).
+- L10-13, the summary: "the lead-only move is" (not blocked by the test).
 - L178-182, section 3 ("Options"): "the one that would fail the
   effort-policy test".
 - L346-348, section 7 (the hardcode-inventory paragraph): "neither is
--- a/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
+++ b/docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md
@@ -47,19 +47,19 @@ blocker that is now cleared. Line numbers are current as of this commit; a
 document that edits itself drifts, so each entry also carries a section
 heading and a quoted phrase to `grep` for if the number is wrong again:
 
-- L10-13, the summary: "the lead-only move is not" (blocked by the test).
+- L10-13, the summary: "the lead-only move is" (blocked by the test).
 - L178-182, section 3 ("Options"): "the one that would fail the
   effort-policy test".
 - L346-348, section 7 (the hardcode-inventory paragraph): "neither is
   blocked by the effort-policy blocker below".
 - L350-360, section 7, the "Flag prominently" block: "blocked by this
-  repo's own test today".
+  repo's own".
 - L362-368, section 7, the alias-table paragraph: "the same
   `shared/models.md` alias table".
 - L409-412, section 8, the recommendation: "until the effort-policy test
   gains an `opus` tier".
 - L414-417, section 8 body: "This is option (c), not (b)".
-- L502-503, section 9, first extend-trigger: "the effort-policy test gains
+- L502-503, section 9, first extend-trigger: "effort-policy test gains
   an `opus` tier (a follow-up issue, not this package)".
 
 None of that prose is edited by this addendum; it stays as written on
@@ -73,7 +73,7 @@ first extend-trigger names the same condition: "the effort-policy test gains
 an `opus` tier (a follow-up issue, not this package)." Both are written as
 satisfied-condition triggers, and both conditions are now met. Meeting them
 does not license moving architect, reviewer, or the critics to Opus 5. The
-effort-policy test was only ever the mechanical blocker (section 8's second
+effort-policy test was only ever the mechanical blocker (section 8's first
 leg, below); the capability reason (section 6, unchanged and unmet) was
 always the independent, load-bearing reason for the hold, and clearing a
 test does not clear a capability question no first-party comparison has


sv-tmueller:
## Tester verdict, round 1: PASS

Commit `def8a63`. No findings. Every claim was reproduced rather than accepted.

### Criterion 1, all eight anchor greps

```
"the lead-only move is"                                                    -> 12, 50 (self-hit, expected)
"the one that would fail the effort-policy test"                           -> 180
"neither is blocked by the effort-policy blocker below"                    -> 347
"blocked by this repo's own"                                               -> 356
"the same `shared/models.md` alias table"                                  -> 362
"until the effort-policy test gains an `opus` tier"                         -> 411
"This is option (c), not (b)"                                              -> 414, 61 (self-hit, pre-existing)
"effort-policy test gains an `opus` tier (a follow-up issue, not this package)" -> 503
```

All eight reach their targets, with only the two documented self-hits.

**The old anchors were confirmed genuinely broken**, so this is a real correction rather than a cosmetic one: on `origin/main` the three old anchors straddled L12/L13, L356/L357 and L502/L503 respectively, and none could have matched a single line.

**Line-count preservation verified.** Exactly four one-line replacements, no lines added or removed, and `wc -l` is 548 on both `origin/main` and this branch. No `L` reference anywhere drifted.

**Two wrinkles examined and judged non-defects.** Anchor 1 stays unique only because `grep` is case-sensitive by default: `grep -i` also returns L417, which has the same phrase capitalized. Since the addendum instructs a plain `grep` and suggests no `-i`, this is fit for purpose. Anchor 8's index entry is still wrapped across two lines, but joining with a single space reconstructs a string that lives entirely on L503 and matches only there, and GitHub's rendered markdown displays a soft break inside a list item as one line regardless.

### Criteria 2 through 6

- **Criterion 2.** Section 8's body states the effort-policy-test blocker before the vendor-capability point, and the addendum's own two-legs paragraph orders them the same way. "First leg" is correct; "second leg" was wrong.
- **Criterion 3.** The message now matches the asserted string exactly, and the assertion line itself is untouched by the diff.
- **Criterion 4.** `criticWithFallback` extracted from all three workflow files and hashed: identical `sha256` across all three. `sliceFnSrc` does include the comment text, so the byte-identity test would fail on divergence. `grep -rn 'proxied' .claude/workflows/` returns nothing.
- **Criterion 5.** Both enumerations now list `log`. The underlying fact was independently confirmed: all three workflow files call `log()` with no local declaration, and `helpers.test.mjs` states that the helper's only free variables are the runtime globals `agent` and `log`, stubbing `log` as an injected global in its sandbox.
- **Criterion 6.** `npm test`: exit 0, 70 tests / 9 suites / 70 pass / 0 fail.

### Non-goals

`git diff origin/main...HEAD --shortstat` is 6 files, 14 insertions, 14 deletions, 8 hunks, all accounted for. Addendum line count unchanged at 548. `opts` unchanged everywhere. The quota-versus-skip comment untouched. The wiring assertion unchanged. No figure, recommendation, or confidence level in the research doc touched.

### One observation for the human, deliberately not fixed here

`docs/plans/37-review-codebase.md:37` also enumerates the ambient globals as `agent`, `parallel`, `phase`, `args` without `log`. The tester judged it out of scope, and I agree: plan docs are point-in-time snapshots by repo convention rather than living references, and this issue's scope names the codebase map specifically. Flagging it so the decision is visible rather than silent. If you want it corrected, it is a one-line follow-up.

### Untested claims

Criteria 1, 2 and 5 are documentation content with no automated coverage in `npm test`, so they rest on the greps and reads reported above rather than on a test. The same applies to the non-goal claims about the research doc's line count, figures, recommendations and confidence levels, which were verified by diff and `wc` rather than by any test.

---

sv-tmueller:
## Reviewer verdict, round 1: CHANGES_REQUESTED

Spec pass clean, one must-fix in the quality pass. The lead independently reproduced the must-fix before acting on it.

### Spec pass: all six criteria met

The reviewer re-ran all eight greps rather than accepting the transcript, and reached the same targets. On the criterion-1 judgment call it was asked to make: shortening a quoted anchor is a legitimate reading, because the addendum's own rule forbids editing the target prose ("None of that prose is edited by this addendum"), so shortening the quote was the only way to satisfy "verbatim on a single source line". Anchor 4 is unique with a single hit. Anchor 1's case-sensitivity reliance is acceptable, since each entry also carries a section label as a second locator, so even a `grep -i` triple hit still finds the target. Anchor 8's wrapped index entry is fine because criterion 1 governs the target line, and other entries wrap identically on `main`.

The clause removal reads as a complete sentence after the cut, not clipped. On `log`'s placement: there is no runtime-documented order anywhere in the repo to match, and the two edited enumerations now agree with each other, which is the consistency that matters. `npm test`: 70 / 9 / 70 pass / 0 fail.

Non-goals respected. The `docs/plans/37-review-codebase.md:37` observation is out of scope, concurring with the tester.

### Finding 1, must-fix: the shortened anchor inverts the entry's meaning

`docs/research/2026-07-24-opus-5-vs-fable-5-judgment-seats.md:50`

On `main` the entry read `"the lead-only move is not" (blocked by the test)`, where the quote plus the parenthetical reconstructed the summary's elided predicate: the move is **not** blocked. This branch removed `not` from the quote to fix the wrap, but left the parenthetical unchanged:

```
- L10-13, the summary: "the lead-only move is" (blocked by the test).
```

That now reads as "the lead-only move **is** blocked by the test", the inverse of its target and of the document's central point. Verified against the source: L12-13 reads "the lead-only move is / not.", and L417 states "The lead-only move is not blocked by that test". No other index entry carries such a gloss, so the continuation reading is the intended one.

A factual-corrections PR turned a meaning-correct but grep-broken entry into a grep-correct but meaning-inverted one. **Fix: change the parenthetical to `(not blocked by the test)`.** That keeps the anchor, the line count, and every grep result unchanged.

### Finding 2, should-fix: a commit body misstates the history

Commit `787b9a2`'s body frames the missing `log` as an original omission repeated twice. The reviewer checked the history and found otherwise: at the map's generating commit (`d06fd07`, 2026-07-05) no workflow script referenced `log` at all, and `log()` entered with `9465e67` (#270, 2026-07-25). The map was accurate on its own date, so this is a sanctioned update of a snapshot that went stale, not the correction of an error the artifact had when written.

**Lead decision: correct this in the PR body rather than by rewriting the commit.** Amending a pushed commit body to fix a framing nuance costs a history rewrite for no functional gain, and this repo squash-merges, so the squash message is where the durable record is written. The PR body is corrected in this revision, and the point is flagged here for the squash message.

This matters beyond pedantry, and the reviewer's framing of why is worth preserving: in-place edits to dated artifacts stay a safe precedent only when they are either corrections of errors the artifact had on its own date, or explicitly labeled as sign-off-sanctioned updates. This one is the second kind, and the record should say so.

### Nits, recorded and not acted on

- Anchor 1 is unique only case-sensitively; `grep -i` also returns L417. Fit for the plain `grep` the addendum recommends, and disambiguated by the section label. No change.
- Anchor 4's quote ends mid-noun-phrase. Unique and functional, awkward only.
- Commit `354a78a` uses `fix:` for a comment-only change; `docs:` or `chore:` fits the taxonomy better. Defensible, since the comment asserted a claim that did not hold. Correctable at squash if you want.

Commit hygiene otherwise clean: all four imperative, lowercase, no trailing period, bodies explain why. No em dashes and no cliche phrasing in the diff.

Dispatching fix round 1 for finding 1 only.

---

sv-tmueller:
## Fix round 1 applied

Commit `f82c27f` on top of `def8a63`. One line, exactly as prescribed:

```diff
-- L10-13, the summary: "the lead-only move is" (blocked by the test).
+- L10-13, the summary: "the lead-only move is" (not blocked by the test).
```

Lead verification: L50 now reads `- L10-13, the summary: "the lead-only move is" (not blocked by the test).`, which agrees with the target at L12-13 ("the lead-only move is / not.") and with L417 ("The lead-only move is not blocked by that test"). The quoted anchor is untouched, its grep still returns L12 plus the L50 self-hit, and the file is still 548 lines, so no `L` reference drifted.

**One correction to my own instruction.** I told the developer to expect `git diff origin/main...HEAD --stat` to show 15 insertions and 15 deletions after this fix, one more than the previous 14/14. That was wrong, and they were right to flag it rather than force the number to match. A three-dot diff compares final tree states, so line 50 was already counted as one insertion and one deletion before this round; changing its content again does not add a second pair. 14/14 is correct and is what the tree shows. The report came back `DONE_WITH_CONCERNS` solely because of my bad expectation, not because of anything in the fix.

Re-testing now, with the tester specifically asked to check whether the other two shortened anchors (the entries at L55-56 and L62-63) have the same gloss-inversion problem, since the round-1 fix shortened three quotes and only one was caught.

---

sv-tmueller:
## Tester verdict, round 2: PASS

Commit `f82c27f`. No findings, and no untested claims remaining.

**The fix is semantically right, not merely different.** L50 now reads `- L10-13, the summary: "the lead-only move is" (not blocked by the test).` The tester traced the elided predicate: L12-13 read "the lead-only move is / not.", where "blocked" carries over from the preceding clause "extending further ... is blocked today by the effort-policy test", and L417 states "The lead-only move is not blocked by that test". Both confirm the corrected gloss. No overcorrection: the quote itself is untouched and only the parenthetical changed.

- **Criterion 1 intact.** All eight anchors re-run, each resolving to its target (L12, L180, L347, L356, L362, L411, L414, L503) with the two expected self-hits at L50 and L61.
- **Line count** still 548, matching `origin/main`, so no `L` reference drifted.
- **Criterion 6.** `npm test`: exit 0, 70 tests / 9 suites / 0 fail, including the byte-identity and wiring-message subtests.
- **Fix scope.** `git diff def8a63..HEAD` touches exactly one line; diffing the codebase map, the three workflow files and the wiring test against `def8a63` is empty, so criteria 2 through 5 are undisturbed.
- **Diff shape** 6 files, 14 insertions and 14 deletions, consistent with L50 being an already-changed line in a tree diff.
- **Non-goals** all still respected, including the confidence level in section 8, unchanged from `main`.

### The failure class, checked across every instance rather than just the one caught

This is what I asked for specifically, because checking one instance of a defect class and not the others is how the second one ships.

- **Anchors 4 and 8**, the other two shortened in round 1: both are plain substring quotes with no parenthetical gloss reconstructing an elided predicate. Each was shortened only to fit a single source line (L356 and L503) and still asserts exactly what its target says. Neither carries anchor 1's failure class. Confirmed individually.
- **The five untouched entries**: all plain quotes with no gloss, so none has the reconstructed-negation pattern, either on `main` or here. No pre-existing mismatch that this PR should arguably have caught.
- **Surrounding prose** at L44-49 and L65-73, including the full un-shortened restatement of the L502-503 quote, is consistent with the corrected entry. No contradiction introduced.

Re-dispatching the reviewer, since the round-1 verdict was CHANGES_REQUESTED and a package needs a reviewer APPROVE on record to be done. (This differs from #273 in the same batch, where the reviewer had already given APPROVE and the follow-up edits were its own dictated wording, so no re-review was needed there.)

---

sv-tmueller:
## Reviewer verdict, round 2: APPROVE

No must-fix findings remain. The package is clear to ship.

**The must-fix is discharged**, and the reviewer checked the grammar behind the fix before ruling on it rather than accepting the reading it was handed. L10-13 reads "One blocker: extending further to architect/reviewer/critics is blocked today by the effort-policy test having no `opus` tier (section 7); the lead-only move is not." The "is not" is verb-phrase ellipsis whose only grammatical antecedent is the preceding predicate "blocked today by the effort-policy test having no `opus` tier". The alternative parse, "is not [extending further]", fails because "extending further" is the prior clause's subject rather than its predicate, and there is no verb form after "is" to license it. L417 states the claim unelided and removes any doubt. So the corrected gloss is a faithful compression, not an overcorrection into a new error.

That mattered: had the reading been wrong, the fix would have been confidently wrong in a fresh way, and this was the last gate before shipping.

**No collateral disturbance.** Line count 548 on both sides, all eight anchors resolve, tree diff unchanged at 6 files / 14 / 14, `npm test` exit 0.

**Finding 2's resolution accepted.** The reviewer does not require the commit body amended: the repo squash-merges, so `787b9a2`'s body never reaches `main`, the durable records are the PR body and the squash message, and rewriting a pushed branch mid-review would invalidate the tester's verified SHA and force a re-verification round for no functional gain. It was a should-fix, and should-fix items belong in the PR for human review, which is where it landed.

One honest caveat the reviewer recorded: with no `gh` and no GitHub tools it could not read the PR body itself, so that ruling rests on the represented contents rather than on direct inspection. The corrected framing is in the PR body above, naming `d06fd07`/2026-07-05 and `9465e67`/#270/2026-07-25.

### For you at merge time, the one thing that needs a human hand

GitHub's squash message defaults to the concatenated branch-commit bodies, which would carry `787b9a2`'s incorrect "the same omission twice" framing into `main` despite the PR body being right. **Please replace the default squash message** with the corrected framing: the `log` addition is a sign-off-sanctioned update of a snapshot that went stale at #270, not the correction of an omission the map had when written. Commit `354a78a`'s `fix:` type is also better as `docs:` if you are editing the message anyway.

### Nits carried forward, unactioned by design

- Anchor 1's uniqueness is case-sensitive; a case-insensitive grep also hits L417. Fit for the plain `grep` the addendum recommends.
- Anchor 4's quote ends mid-noun-phrase. Cosmetic only; the anchor resolves.
- `docs/plans/37-review-codebase.md:37` also lists the ambient globals without `log`. Out of scope, one-line follow-up if wanted.

### Pipeline complete

Last tester verdict PASS, reviewer APPROVE, one review fix round. Marking this PR ready for review.

---

PR #282, opened 2026-07-26
docs: make the Fable-quota Opus override a routing rule

### LINKED ISSUE #273: Make the Fable-quota Opus override a routing rule, not a remembered procedure

sv-tmueller:
Closes #273

Part of batch #278. Closes the dispatchable remainder of #261.

## Why

#267 automated recovery for the workflow critic stages, which was the half that mattered most because workflow scripts run unattended. Lead-level dispatches still depended on the lead noticing a quota death and remembering to re-dispatch that one agent with a per-call Opus override, from prose in the Model policy.

The evidence that remembering is not enough is substantial: across two consecutive batches the Fable 5 limit was hit seven times on four separate occasions, and in the most recent batch every judgment dispatch died on quota and needed the override applied by hand. A fresh lead session should apply the response without rediscovering it, and the model switch should be recorded rather than silent.

## What changed

Three files, docs only, 18 insertions and 4 deletions. All three use the same phrases ("per-call Opus override", "the Agent tool's `model` param") so grep finds them together.

**`.claude/skills/tm-kickoff/SKILL.md`** gains a routing rule, placed immediately before the "Never re-dispatch an unchanged prompt" bullet because it has to define its relationship to that one. The rule names the trigger (an architect or reviewer dispatch dying on Fable 5 quota, either the limit error or an empty return while Fable is exhausted), prescribes the response (re-dispatch that one agent with the same task and a per-call Opus override), rules out the wrong responses (no frontmatter pin flip, no session-model flip, the agent's own effort pin still applies), requires the switch to be logged as a decision comment, and scopes itself to lead dispatches since workflow critic stages recover on their own via `criticWithFallback`.

The clause "the model override is the change that permits the re-dispatch" is deliberate: without it the new rule would appear to contradict the "never re-dispatch an unchanged prompt" bullet that follows.

**`.claude/skills/tm-advisor/SKILL.md`** gains one sentence on the "In-scope decisions are yours" bullet, settling two things the kickoff rule does not: that a Fable-quota death is an in-scope decision rather than grounds to park, and that inside a batch the log entry goes on the batch issue. It defers the mechanics to kickoff's rule rather than restating them, so the two cannot drift apart.

**`.claude/team-guide.md`** reconciles the Model policy's Fallback bullet instead of leaving it saying something different. The sentence that previously described the single-agent override becomes a cross-reference naming the kickoff rules as owner. The opening parenthetical "(a procedure, not automatic)" becomes "(the lead-session flip is a manual procedure; the workflow critics and single-agent dispatches recover by rule)", because after this change the old wording would misdescribe two of the bullet's three paths.

## The distinction this had to preserve

The Model policy bullet describes three separate mechanisms, and blurring them would be the easy mistake here:

1. **The lead-session model flip** for a longer outage. Stays a manual procedure, unchanged.
2. **The workflow critic auto-retry** from #267 (`criticWithFallback`). Unchanged, and the new rule explicitly says it does not cover this.
3. **The single-agent per-call override.** The only one moving under the new routing rule.

Where the log entry goes differs by pipeline, so the rule covers both: a comment on the package issue under a plain `/tm-kickoff` run, and the batch issue inside a `/tm-advisor` batch. The package issue was chosen for the plain case because the architect stage, the seat most likely to die, runs before any PR exists, while the package issue exists at every stage.

## Held to the non-goals

Fable 5 stays the first-tried judgment model. No `effort:` pin changed and no agent frontmatter edited. No batch-level "Fable is down" switch. Sonnet is not introduced as a judgment fallback. #261's plugin-cache-overwrite problem is untouched and still needs its own design pass.

Four near-miss sites were checked and deliberately left alone because they remain correct: the `architect.md` and `reviewer.md` frontmatter comments (and frontmatter edits are a non-goal anyway), `tm-map-codebase.js`'s pointer to team-guide for the lead-level story, `effort-policy.test.mjs`'s header describing the pin-flip path, and the dated 2026-07-24 research doc.

## Verification

- `npm test` -> exit 0, 70 tests / 9 suites / 0 fail. No test reads any SKILL.md or team-guide.md, so this criterion is verify-only; `effort-policy.test.mjs` reads agent frontmatters and workflow JS, `ci-template-policy.test.mjs` reads the tm-new-project CI template, and `helpers.test.mjs` reads workflow JS.
- `git diff --stat` is exactly the three files, with nothing under `.claude/agents/` or `.claude/workflows/`.
- `grep -rn 'per-call Opus override' .claude/ docs/ CLAUDE.md` returns exactly four hits: the three policy sites plus the dated research doc, which was on the do-not-touch list. The three policy sites agree with each other, with kickoff owning the mechanics and the other two pointing at it.
- Read-through check: a fresh lead reading only the kickoff routing rules has the trigger, the response, what not to do, the effort-pin note, the logging requirement, and the relationship to both neighbouring rules, without needing to open team-guide.md.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tk5nzbKYpBKsRVzkXaNi7S)_

### DIFF
--- a/.claude/skills/tm-advisor/SKILL.md
+++ b/.claude/skills/tm-advisor/SKILL.md
@@ -163,6 +163,9 @@ advisor differs from a plain `/tm-kickoff` run in three ways:
   scope and acceptance criteria (a NEEDS_DECISION, an arbitration outcome,
   an interpretation call), decide it and post the decision with its reasoning
   as a comment on the batch issue before acting on it.
+  A Fable-quota death on a judgment-seat dispatch is such a decision:
+  apply kickoff's per-call Opus override routing rule and log the switch
+  on the batch issue. It is not grounds to park.
 - **Narrower parking gate.** Park (swap `in-progress` for `needs-human`) only
   for: a change to scope or acceptance criteria, a new dependency or cost,
   anything irreversible or outward-facing, or a conflict with
--- a/.claude/skills/tm-kickoff/SKILL.md
+++ b/.claude/skills/tm-kickoff/SKILL.md
@@ -121,15 +121,15 @@ Routing rules:
   do not re-raise it.
 - If the architect's sub-plan says the work exceeds the size label, stop
   that package and report it (re-label and split per CLAUDE.md "Sizing").
-- An architect or reviewer dispatch dies on Fable 5 quota (the limit
-  error, or an empty return while Fable is exhausted): re-dispatch that
-  one agent with the same task and a per-call Opus override (the Agent
-  tool's `model` param). Do not flip a frontmatter pin or the session
-  model; the agent's own effort pin (xhigh) still applies. The model
-  override is the change that permits the re-dispatch. Log the switch as
-  a decision comment on the package issue; a model switch is never
-  silent. Workflow critic stages recover on their own
-  (criticWithFallback); this rule covers lead dispatches only.
+- An architect or reviewer dispatch dies on Fable 5 quota (the limit error,
+  or an empty return while Fable is exhausted): re-dispatch that one agent
+  with the same task and a per-call Opus override (the Agent tool's `model`
+  param, value `opus`). Do not flip a frontmatter pin or the session model;
+  the agent's own effort pin (xhigh) still applies. The model override is
+  the change that permits the re-dispatch. Log the switch as a decision
+  comment on the package issue; a model switch is never silent. Workflow
+  critic stages recover on their own (criticWithFallback); this rule covers
+  lead dispatches only.
 - Never re-dispatch an unchanged prompt; something in the task must change
   first.
 - Cap: 3 fix rounds per stage, counted from the PR comments. Tester and
--- a/.claude/skills/tm-kickoff/SKILL.md
+++ b/.claude/skills/tm-kickoff/SKILL.md
@@ -121,6 +121,15 @@ Routing rules:
   do not re-raise it.
 - If the architect's sub-plan says the work exceeds the size label, stop
   that package and report it (re-label and split per CLAUDE.md "Sizing").
+- An architect or reviewer dispatch dies on Fable 5 quota (the limit
+  error, or an empty return while Fable is exhausted): re-dispatch that
+  one agent with the same task and a per-call Opus override (the Agent
+  tool's `model` param). Do not flip a frontmatter pin or the session
+  model; the agent's own effort pin (xhigh) still applies. The model
+  override is the change that permits the re-dispatch. Log the switch as
+  a decision comment on the package issue; a model switch is never
+  silent. Workflow critic stages recover on their own
+  (criticWithFallback); this rule covers lead dispatches only.
 - Never re-dispatch an unchanged prompt; something in the task must change
   first.
 - Cap: 3 fix rounds per stage, counted from the PR comments. Tester and
--- a/.claude/team-guide.md
+++ b/.claude/team-guide.md
@@ -192,8 +192,9 @@ Rationale: docs/team-guide-rationale.md.
   lead stays on the bounded tm- machinery. Fable costs 2x Opus 5 per
   token. The premium is bounded in aggregate, not per batch
   (docs/research/2026-07-06-token-burn-investigation.md, driver 3).
-- Fallback: Opus 5 at xhigh effort (a procedure, not automatic). When
-  Fable 5 is unavailable, rate-limited, quota-exhausted, or
+- Fallback: Opus 5 at xhigh effort (the lead-session flip is a manual
+  procedure; the workflow critics and single-agent dispatches recover by
+  rule). When Fable 5 is unavailable, rate-limited, quota-exhausted, or
   refuses the workload, switch the lead with `/model claude-opus-5`; for a
   longer outage flip the `fable` pins to `opus` in the `architect` and
   `reviewer` frontmatters. The Fable-critic stage of every `tm-` workflow
@@ -201,8 +202,9 @@ Rationale: docs/team-guide-rationale.md.
   returns nothing, so no flip is needed there, though each run still burns
   one doomed Fable dispatch before the retry fires. Flip back when Fable
   returns. For a single role-agent dispatch hitting Fable quota mid-batch,
-  dispatch that one agent with a per-call Opus override (the Agent tool's
-  `model` param) instead of flipping every pin. The ladder is Fable -> Opus;
+  the kickoff routing rules own the response: a per-call Opus override
+  (the Agent tool's `model` param), logged as a decision, no pin flip
+  (`.claude/skills/tm-kickoff/SKILL.md`). The ladder is Fable -> Opus;
   Sonnet is never a fallback
   for a judgment seat, and only steps in, flagged, if Opus is also down, with
   the review re-run on a judgment model afterward.
--- a/.claude/team-guide.md
+++ b/.claude/team-guide.md
@@ -193,21 +193,20 @@ Rationale: docs/team-guide-rationale.md.
   token. The premium is bounded in aggregate, not per batch
   (docs/research/2026-07-06-token-burn-investigation.md, driver 3).
 - Fallback: Opus 5 at xhigh effort (the lead-session flip is a manual
-  procedure; the workflow critics and single-agent dispatches recover by
-  rule). When Fable 5 is unavailable, rate-limited, quota-exhausted, or
-  refuses the workload, switch the lead with `/model claude-opus-5`; for a
-  longer outage flip the `fable` pins to `opus` in the `architect` and
-  `reviewer` frontmatters. The Fable-critic stage of every `tm-` workflow
-  (`.claude/workflows/`) already retries on opus automatically when Fable
-  returns nothing, so no flip is needed there, though each run still burns
-  one doomed Fable dispatch before the retry fires. Flip back when Fable
-  returns. For a single role-agent dispatch hitting Fable quota mid-batch,
-  the kickoff routing rules own the response: a per-call Opus override
-  (the Agent tool's `model` param), logged as a decision, no pin flip
-  (`.claude/skills/tm-kickoff/SKILL.md`). The ladder is Fable -> Opus;
-  Sonnet is never a fallback
-  for a judgment seat, and only steps in, flagged, if Opus is also down, with
-  the review re-run on a judgment model afterward.
+  procedure; workflow critics recover automatically, single-agent dispatches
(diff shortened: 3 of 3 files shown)


sv-tmueller:
## Tester verdict, round 1: PASS (1 non-blocking finding)

Commit `534efa8`. All five acceptance criteria met. `git diff origin/main...HEAD --stat` is exactly the three claimed files, 18 insertions and 4 deletions, nothing under `.claude/agents/` or `.claude/workflows/`, no `effort:` line touched.

- **Criterion 5.** `npm test`: exit 0, 70 tests / 9 suites / 70 pass / 0 fail. The claim that no test reads SKILL.md or team-guide.md was verified by grep: no hits. The one related hit builds a path array pointing at the tm-new-project CI template, matching `'skills'` as a literal array element rather than a path string. So criterion 5 is not fragile with respect to this PR's prose changes.
- **Criterion 1.** The new bullet names both failure signatures (the limit error, or an empty return while Fable is exhausted) and which seats, and prescribes the exact response. Line widths of 66-75 characters match the surrounding bullets.
- **Criterion 2.** Logging is required, not suggested: "Log the switch as a decision comment on the package issue; a model switch is never silent." Imperative and unconditional.
- **Criterion 3.** The batch-issue versus package-issue difference is not a contradiction. The advisor bullet already establishes that all in-scope decisions are logged on the batch issue, and kickoff's own final routing rule mirrors lead decisions there inside a batch, so plain kickoff logs on the package issue and advisor-mode kickoff mirrors it up. The "Narrower parking gate" bullet's four categories do not cover a Fable-quota death, so "not grounds to park" does not conflict with it. Neither file implies flipping a pin for a single-agent death; both forbid it.
- **Criterion 4.** All three mechanisms remain distinguishable: the lead-session flip stays manual with its `/model` and pin-flip instructions, the critic auto-retry prose is unchanged and cross-referenced, and the single-agent override became a cross-reference instead of restated prose.

### Break-it results, all clean

- No fourth site describes the old remembered-procedure behavior as current policy. Grepped `model param`, `flipping every pin`, `quota`, `Fable 5 limit`, `claude-opus-5` and `modelFallback`; every other hit is unrelated or a historical record consistent with the new rule.
- The rule's claim that workflow critic stages recover on their own was verified against the code: `criticWithFallback` retries once with `{ ...opts, model: 'opus' }` on a null return and throws only if both fail. Genuinely automatic, no lead action.
- `architect.md` and `reviewer.md` both pin `effort: xhigh`, so the "(xhigh)" parenthetical states the right value.
- Frontmatter enumeration checked across all seven agents: architect and reviewer are the only two `model: fable` seats, so the rule's "An architect or reviewer dispatch" is complete and misses no seat.

### Finding 1, non-blocking: the new parenthetical overstates automation

`.claude/team-guide.md:196-197`. The parenthetical "(the lead-session flip is a manual procedure; the workflow critics and single-agent dispatches recover by rule)" pairs a fully automatic mechanism (`criticWithFallback`, zero human action) with one that still requires the lead to notice the quota death and re-dispatch. The rule fixes what to do once noticed, not the noticing. A reader skimming only the parenthetical could conclude the single-agent case is as automatic as the critic case.

Not a factual contradiction, since the following sentences correctly attribute the single-agent case to the kickoff routing rules, which the lead applies. Forwarded to the reviewer to judge whether it rises to must-fix, since the text is this PR's own and precision about exactly this three-way distinction is what criterion 4 is about.

### Untested claims

The entire substance of criteria 1 through 4 is verified by reading only; `npm test` parses nothing in SKILL.md or team-guide.md. A future edit could silently reintroduce the contradiction this PR fixes with no test catching it. The same applies to the non-goals "Fable 5 stays the default judgment model" and "no Sonnet fallback introduced".

---

sv-tmueller:
## Reviewer verdict, round 1: APPROVE (1 should-fix, 1 nit)

All five criteria met, three hunks each mapping to a criterion, no tests, frontmatter, or effort pins touched.

### The four judgment calls the reviewer was asked to settle

1. **The kickoff bullet does execute standalone.** A fresh lead reading only the routing rules gets the trigger with both failure signatures, the affected seats, the action, the prohibitions, the effort clarification, the logging obligation, and the scope boundary against `criticWithFallback`, without opening team-guide.md. That is the issue's stated purpose, so this was the test that mattered most.
2. **The reconciliation with "Never re-dispatch an unchanged prompt" works.** The clause claiming the exception sits directly before that bullet. The reviewer noted a letter-level mismatch (the never-bullet says "something in the task must change", and the model param is not part of the task) but judged it harmless: a reader reaches the never-bullet only after passing the quota rule that pre-empts it, and the never-bullet's purpose, not repeating a failed prompt hoping for a different result, does not apply to a dispatch that never executed.
3. **The layering direction is sound.** team-guide.md already delegates operational detail to skill files by name, and the issue prescribes this direction. The policy bullet still states the substance inline rather than being a bare pointer, so it degrades gracefully if the skill file moves.
4. **The advisor sentence is on the right bullet.** It prescribes a positive action, which is the in-scope-decisions bullet's job, while the parking gate is a closed "only for:" list that already excludes quota deaths by construction.

Commit hygiene clean: all three `docs:`, matching repo precedent, imperative, lowercase, no period, bodies explain why. No em dashes or cliche phrasing, wrap widths match neighbours.

### Finding 1, should-fix: the parenthetical collapses the distinction the prose maintains

`.claude/team-guide.md:195-197`. "(the lead-session flip is a manual procedure; the workflow critics and single-agent dispatches recover by rule)" pairs a fully automatic code path with one that needs the lead to notice the death and re-dispatch. The reviewer judged it should-fix rather than must-fix because it prescribes no action, conflicts with no instruction, and the three sentences after it attribute each mechanism correctly.

### Finding 2, nit: the override's value is never named

`.claude/skills/tm-kickoff/SKILL.md:126-127`. "a per-call Opus override (the Agent tool's `model` param)" says which parameter but not what to pass.

### Lead decision: apply both

Shipping as-is would be defensible, since the reviewer approved and both are optional. Applying them anyway, for two reasons specific to this package:

- Its entire deliverable is precision of wording about a three-way distinction the issue explicitly warns must not be blurred. Leaving a known imprecision in the one sentence that draws that distinction is a poor trade for saving a fix round, particularly in a file the lead executes rather than merely reads.
- The nit goes directly to the issue's stated purpose, that a fresh lead applies the rule "without rediscovering it". A rule that names the parameter but not its value still requires one lookup. Naming `opus` removes it.

Both are one-line edits, and the reviewer supplied exact replacement wording, so there is no design work left.

**Verification plan, and a deliberate deviation from the pipeline.** After the edits I will run a focused re-test (the two changed lines read correctly, `npm test` still green, diff still confined to the three files) but will NOT re-dispatch the reviewer. Re-reviewing wording the reviewer itself dictated verbatim adds nothing, and the change carries no logic. Recording the deviation rather than passing it off as the standard flow.

---

sv-tmueller:
## Tester verdict, round 2 (focused re-test): PASS

Commit `53faed5`. No findings.

- **`npm test`**: exit 0, 70 tests / 9 suites / 0 fail.
- **Diff confinement**: exactly the three files, nothing under `.claude/agents/` or `.claude/workflows/`, no `effort:` line touched.
- **`.claude/skills/tm-advisor/SKILL.md` untouched** by this fix round, confirmed empty diff against `534efa8`.

### The re-wrap smuggled nothing in

This was the round's main risk, since a re-wrap makes a line-based diff hard to read. Checked word by word on whitespace-collapsed text:

- Fallback bullet, `534efa8` to `HEAD`: the only word-level delta is the parenthetical rewording. Everything else, including the "Sonnet is never a fallback for a judgment seat" sentence, is identical apart from line-break positions.
- Kickoff bullet, `534efa8` to `HEAD`: the only word-level delta is `param).` becoming `param, value \`opus\`).`
- Fallback bullet, `origin/main` to `HEAD`: the full two-round delta is contained to the new three-part parenthetical and the single-agent sentence becoming a cross-reference. No other words changed.

That matches the lead's own independent check before dispatch.

### The value `opus` is correct

Verified against `criticWithFallback` in `.claude/workflows/tm-review-changes.js`, which retries with `{ ...opts, model: 'opus' }`, a bare alias rather than `claude-opus-5`, and against the frontmatter convention (`model: fable`, `model: sonnet`). Had the right value been `claude-opus-5`, the nit fix would have been actively misleading.

### Final wording

Parenthetical: "(the lead-session flip is a manual procedure; workflow critics recover automatically, single-agent dispatches by the kickoff routing rule)"

Override phrase: "(the Agent tool's `model` param, value `opus`)"

The three mechanisms are distinguishable, and the kickoff bullet is self-contained and executable without further lookup, which is the issue's stated purpose. Line widths within each file's normal prose range (77 and 75 characters). No em dashes or cliche phrasing in any changed line.

### Pipeline complete

Reviewer APPROVE on round 1, last tester verdict PASS, one fix round applying the reviewer's own should-fix and nit. No re-review, per the decision logged on #278: re-reviewing wording the reviewer dictated verbatim, on a change with no logic and with content-equivalence verified word by word, adds nothing. Marking this PR ready for review.

---

